PR #18, opened 2017-07-01
Add one-sided derivatives at the boundaries

anriseth:
This adds support for leaky BCs.

I can add some tests if you want

### DIFF
--- a/src/fornberg.jl
+++ b/src/fornberg.jl
@@ -44,8 +44,8 @@ immutable LinearOperator{T<:Real,S<:SVector,LBC,RBC} <: AbstractLinearOperator{T
         stencil_coefs        = convert(SVector{stencil_length, T}, calculate_weights(derivative_order, zero(T),
                                grid_step .* collect(-div(stencil_length,2) : 1 : div(stencil_length,2))))
 
-        l_fact = initialize_left_boundary!(low_boundary_coefs,high_boundary_coefs,stencil_length,derivative_order,grid_step,boundary_length,boundary_point_count,LBC,stencil_coefs)
-        r_fact = initialize_right_boundary!(low_boundary_coefs,high_boundary_coefs,stencil_length,derivative_order,grid_step,boundary_length,boundary_point_count,RBC,stencil_coefs)
+        l_fact = initialize_left_boundary!(low_boundary_coefs,high_boundary_coefs,stencil_length,derivative_order,grid_step,boundary_length,boundary_point_count,stencil_coefs,LBC)
+        r_fact = initialize_right_boundary!(low_boundary_coefs,high_boundary_coefs,stencil_length,derivative_order,grid_step,boundary_length,boundary_point_count,stencil_coefs,RBC)
 
         boundary_fn = (l_fact*bndry_fn[1], r_fact*bndry_fn[2])
 
@@ -65,7 +65,7 @@ end
 
 function initialize_left_boundary!{T}(low_boundary_coefs,high_boundary_coefs,stencil_length,
                                    derivative_order,grid_step::T,boundary_length,
-                                   boundary_point_count,LBC=:Neumann,stencil_coefs)
+                                   boundary_point_count,stencil_coefs,LBC=:Neumann)
     high_temp            = zeros(T,boundary_length)
     flag                 = derivative_order*boundary_point_count%2
     aorder               = boundary_length - 1
@@ -82,7 +82,6 @@ function initialize_left_boundary!{T}(low_boundary_coefs,high_boundary_coefs,ste
     # scale!(original_coeffs, first_order_coeffs[end]/original_coeffs[end])
     @. original_coeffs = original_coeffs - first_order_coeffs
     # copy!(first_order_coeffs, first_order_coeffs[1:end-1])
-    push!(low_boundary_coefs, original_coeffs[1:end-1])
 
     if LBC == :None
         for i in 1 : boundary_point_count
@@ -97,6 +96,7 @@ function initialize_left_boundary!{T}(low_boundary_coefs,high_boundary_coefs,ste
             end
         end
     else
+        push!(low_boundary_coefs, original_coeffs[1:end-1])
         for i in 2 : boundary_point_count
             #=  this means that a stencil will suffice ie. we dont't need to worry about the boundary point
             being considered in the low_boundary_coefs
@@ -108,7 +108,7 @@ function initialize_left_boundary!{T}(low_boundary_coefs,high_boundary_coefs,ste
                 pos=i-2
                 push!(low_boundary_coefs, append!([0],calculate_weights(derivative_order, pos*grid_step, collect(zero(T) : grid_step : (boundary_length-1) * grid_step))))
             end
-        end    
+        end
     end
 
     return l_diff
@@ -117,7 +117,7 @@ end
 
 function initialize_right_boundary!{T}(low_boundary_coefs,high_boundary_coefs,stencil_length,
                                    derivative_order,grid_step::T,boundary_length,
-                                   boundary_point_count,RBC=:Neumann)
+                                   boundary_point_count,stencil_coefs,RBC=:Neumann)
     # high_temp            = zeros(T,boundary_length)
     flag                 = derivative_order*boundary_point_count%2
     aorder               = boundary_length - 1


src/boundary_operators.jl
```diff
@@ -22,6 +22,35 @@ function rem1(idx,L)
 end
 
 
+#= LEAKY BCs =#
```
ChrisRackauckas: Purely style remark: Leaky should come after the other ones, down here:

https://github.com/anriseth/PDEOperators.jl/blob/1879f32ced709fd4970d1a72dee9e1710e5553ce/src/boundary_operators.jl#L94

ChrisRackauckas: Also, left and right should be split like the others. This is the only one which isn't following the format.

ChrisRackauckas:
Where does the user constants come in?

anriseth:
> Where does the user constants come in? 

I don't understand what you mean?

ChrisRackauckas:
Could you explain the implementation a little bit? I'm not entirely sure how it's working, where user constants come in for the mix between the derivative and constant values (are there any?), or if we're using the term leaky differently. I've always used leaky == Robin, and this doesn't look the same.

anriseth:
> I've always used leaky == Robin, and this doesn't look the same.

Aah, haha then I've just been stupid and got confused with the name. I wanted one-sided finite-differences at the boundaries, which didn't seem to be supported.

ChrisRackauckas:
>Aah, haha then I've just been stupid and got confused with the name. I wanted one-sided finite-differences at the boundaries, which didn't seem to be supported.

Hmm that's interesting and maybe should be added under a difference name (`:none`?), but what's the use case?

anriseth:
Maybe `LBC = :None` is what I'm after

anriseth:
The use case is that I'm solving PDEs where for example information travels from the interior to one of the boundaries, so there is no need to specify BCs there

src/boundary_operators.jl
```diff
@@ -146,8 +146,8 @@ function convolve_BC_right!{T<:Real,S<:SVector,LBC}(x_temp::AbstractVector{T}, x
 end
 
 
-function convolve_BC_right!{T<:Real,S<:SVector,LBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,LBC,:Leaky})
-    Threads.@threads for i in 1 : A.boundary_point_count
+function convolve_BC_right!{T<:Real,S<:SVector,LBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,LBC,:None})
+    Threads.@threads for i in 1 : A.bouorndary_point_count
```
ChrisRackauckas: typo

src/boundary_operators.jl
```diff
@@ -65,6 +65,21 @@ function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x:
 end
 
 
+function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,:Leaky,RBC})
+    Threads.@threads for i in 1 : A.boundary_point_count
+        @inbounds for i in 1 : A.boundary_point_count
```
ChrisRackauckas: should be one loop on i here

src/boundary_operators.jl
```diff
@@ -131,6 +146,21 @@ function convolve_BC_right!{T<:Real,S<:SVector,LBC}(x_temp::AbstractVector{T}, x
 end
 
 
+function convolve_BC_right!{T<:Real,S<:SVector,LBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,LBC,:Leaky})
+    Threads.@threads for i in 1 : A.boundary_point_count
+        @inbounds for i in 1 : A.boundary_point_count
```
ChrisRackauckas: one loop on i

src/boundary_operators.jl
```diff
@@ -65,6 +65,21 @@ function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x:
 end
 
 
+function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,:None,RBC})
+    Threads.@threads for i in 1 : A.boundary_point_count
+        @inbounds for i in 1 : A.boundary_point_count
+            bc = A.low_boundary_coefs[i]
+            tmp = zero(T)
+            startid = max(0,i-1-div(A.stencil_length, 2))
```
ChrisRackauckas: this one line I don't understand. Other than that, looks good.

ChrisRackauckas:
Yeah, this'll need tests for sure. Not sure what the best test is other than for it to compute a `full` using matrix vector multiplication and checking the values near the boundaries.

src/boundary_operators.jl
anriseth: It takes into account that we don't use the boundary node in the cases under fornberg.jl:109. The issue here is that A.boundary_point_count is too large for what `BC=:None` needs.

src/boundary_operators.jl
```diff
@@ -65,6 +65,19 @@ function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x:
 end
 
 
+function convolve_BC_left!{T<:Real,S<:SVector,RBC}(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::LinearOperator{T,S,:None,RBC})
+    Threads.@threads for i in 1 : A.boundary_point_count
+        @inbounds bc = A.low_boundary_coefs[i]
+        tmp = zero(T)
+        startid = max(0,i-1-div(A.stencil_length, 2))
```
shivin9: maybe you can stash away `div(A.stencil_length, 2)` into variable at the start of loop

shivin9:
I have pushed the changes where I broke the `initialize_boundaries` function into two separate functions as it was getting really messy. Can you please update the `:none` boundary condition accordingly.

src/fornberg.jl
```diff
@@ -102,6 +103,30 @@ function initialize_boundaries!{T}(low_boundary_coefs,high_boundary_coefs,
     end
 
     for i in 1 : boundary_point_count
+        if LBC == :None
+            # One-sided stencils require more points for same approximation order
+            # TODO: I don't know if this is the correct stencil length for i > 1?
+            boundary_stencil_length = derivative_order+approximation_order
```
shivin9: this is same as the `boundary_length` variable

src/fornberg.jl
```diff
@@ -55,23 +55,24 @@ immutable LinearOperator{T<:Real,S<:SVector,LBC,RBC} <: AbstractLinearOperator{T
...
 
 
 function initialize_boundaries!{T}(low_boundary_coefs,high_boundary_coefs,
-                                   derivative_order,grid_step::T,boundary_length,
-                                   boundary_point_count,LBC,RBC)
+                                   derivative_order, approximation_order, grid_step::T,boundary_length,
+                                   boundary_point_count,LBC,RBC, stencil_coefs)
     high_temp            = zeros(T,boundary_length)
     flag                 = derivative_order*boundary_point_count%2
     aorder               = boundary_length - 1
```
shivin9: `aorder` is same as `approximation_order`, you can use this directly instead of getting it from function definition.

shivin9:
@anriseth maybe what you want was our original [implementation](https://github.com/shivin9/PDEOperators.jl/blob/master/src/fornberg.jl#L41). The commented part generates exactly what you want. 
``` julia
julia> A = LinearOperator{Float64}(2,2,1/99,100,:Neumann,:Neumann;bndry_fn=(-1,1));
julia> A.low_boundary_coefs
3-element Array{Array{Float64,1},1}:
 [-3.5, 4.0, -0.5]          
 [0.0, 2.0, -5.0, 4.0, -1.0]
 [0.0, 1.0, -2.0, 1.0]      
```
The third stencil does not take into account the boundary point and is same as the interior stencil `[1, -2, 1]` just with one extra `0.0`.

src/boundary_operators.jl
ChrisRackauckas: For reference that "stash away" is referred to as "hoisting". I think LLVM might hoist it out of the loop, but yes it's better to be safe.

shivin9:
Well just asking for an opinion, do you think that `boundary_point_count` should be `stencil_length - div(stencil_length,2) - 1` instead of `stencil_length - div(stencil_length,2) + 1` as it is right now?
This is because in the `A.low_boundary_coefs` the interior stencil does appear once ie. at the second last position before we use the one-sided stencil for the last (most interior) point.
``` julia
julia> A = LinearOperator{Float64}(3,2,1.0,100,:None,:None);
julia> A.low_boundary_coefs
4-element Array{Array{Float64,1},1}:
 [-2.5, 9.0, -12.0, 7.0, -1.5]
 [-1.5, 5.0, -6.0, 3.0, -0.5] 
 [-0.5, 1.0, 0.0, -1.0, 0.5]  
 [0.5, -3.0, 6.0, -5.0, 1.5]  
julia> A.stencil_coefs
5-element SVector{5,Float64}:
 -0.5
  1.0
  0.0
 -1.0
  0.5
```
It seems totally valid to just exclude the last two points from boundary points.

anriseth:
> Well just asking for an opinion, do you think that boundary_point_count should be stencil_length - div(stencil_length,2) - 1 instead of stencil_length - div(stencil_length,2) + 1 as it is right now?

For the cases I have seen, with `:None` BCs, the boundary_point_count should have two fewer points. I am not sure if this generalises to the other BCs or not?

For reference, I get the following lower stencils (you last element seems incorrect?):
```julia
julia> A = LinearOperator{Float64}(3,2,1.0,10,:None,:None)
julia> A.low_boundary_coefs
4-element Array{Array{Float64,1},1}:
 [-2.5, 9.0, -12.0, 7.0, -1.5]
 [-1.5, 5.0, -6.0, 3.0, -0.5] 
 [-0.5, 1.0, 0.0, -1.0, 0.5]  
 [-0.5, 1.0, 0.0, -1.0, 0.5] 
```

shivin9:
Yes it will be that way because that's the one-sided stencil for the last boundary point. The Fornberg algorithm will output that only.

anriseth:
> Yes it will be that way because that's the one-sided stencil for the last boundary point. The Fornberg algorithm will output that only.

In my implementation, this is taken care of to enforce a central differencing for all "boundary points" that are far enough from the boundary. This is the logic of line 93 in fornberg.jl

shivin9:
Thanks!

shivin9:
Sorry that was not merged properly. I have some more API changes, can pull to get changes and then I will merge your PR into new_boundary_conditions

anriseth:
Thank you :smile:

PR #27, opened 2017-08-06
Restructure and Rename

ChrisRackauckas:
This moves the concrete operator types out of DiffEqBase to here, leaving only the interface specification there. Thus this includes DiffEqScalar and DiffEqArrayOperator. The previous operators all share a lot of code, so those are now in a folder. All of those operators now fall under the abstract type `AbstractDiffEqDerivativeOperator <: AbstractDiffEqLinearOperator` as they are linear operators which describe discretizations of derivatives.

This then does the rename to DiffEqOperators as PDEOperators may be a too "domain-special acronym" to get registered, while we have free reign to register things under DiffEq. It also covers the fact that more than PDEs can be specified by these operators anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ notifications:
 # uncomment the following lines to override the default test script
 #script:
 #  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("PDEOperators"); Pkg.test("PDEOperators"; coverage=true)'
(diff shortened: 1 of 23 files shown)


ChrisRackauckas:
That should be all of it. Tests fail at the same place as they do on master.

abuAnat:
Does DiffEqOperators work for 2D problems?

ChrisRackauckas:
`A*u + u*B`, if `u` is a matrix of the points in 2D space, will have `A` be an operator on the `y` derivative and `B` an operator on the `x` (look at the way it applies the stencil).

PR #30, opened 2017-08-14
Fixed faulty initialization of DiffEqScalar

jagot:
The error showed up when using a `f(t)` type DiffEqScalar.


ChrisRackauckas:
Thanks!

PR #31, opened 2017-08-15
Fix DiffEqScalar:s

jagot:


src/array_operator.jl
```diff
@@ -56,12 +59,12 @@ Base.:*(L::DiffEqArrayOperator,b::AbstractArray) = L.α.coeff*L.A*b
 
 function Base.A_mul_B!(v::AbstractVector,L::DiffEqArrayOperator,b::AbstractVector)
     A_mul_B!(v,L.A,b)
-    scale!(b,L.α.coeff)
+    scale!(v,L.α.coeff)
```
ChrisRackauckas: aha! That's why it wasn't working. Good catch.

ChrisRackauckas:
Thanks!

PR #49, opened 2018-02-20
"FiniteDifference" Operator for irregular spaced grids

fgerick:
Hello there,
I have implemented the generic case for irregular spaced grids. I followed the naming already mentioned in the code: ``FiniteDifference`` , although maybe there is a better name for it. I rearranged some existing functions to differentiate better between functions for the ``AbstractDerivativeOperator`` and the individual ones. I haven't implemented the right Neumann boundary yet, nor the Robin BC. I have tried to think of some tests, but I'm not really sure if these are of any value. This is my very first contribution to an open source project, so forgive me for any rookie mistakes. The tests failing are the ones from the KdV.jl test.


fgerick:
Also there was an error for the ``negate!()`` function used in Neumann BC of the ``DerivativeOperator`` if the ``flag`` value is 1. I fixed it by calling ``eltype(arr[1])`` instead of ``eltype(arr)``. I'm not sure if that causes any problems elsewhere!?

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    v = zeros(T, A.dimension)
+    v[i] = one(T)
+    copy!(v, A*v)
+    return v
+end
+
+
+# UnitRanges
+@inline function getindex(A::AbstractDerivativeOperator, rng::UnitRange{Int}, cc::Colon)
+    m = full(A)
```
ChrisRackauckas: That looks like a memory usage surprise, but for now that's fine.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    scaled appropriately according to the derivative order and the degree of discretizaiton.
+    We also update the time stamp of the DerivativeOperator inside to manage time dependent
+    boundary conditions.
+=#
+function Base.A_mul_B!(x_temp::AbstractVector{T}, A::Union{DerivativeOperator{T},UpwindOperator{T}}, x::AbstractVector{T}) where T<:Real
+    convolve_BC_left!(x_temp, x, A)
+    convolve_interior!(x_temp, x, A)
+    convolve_BC_right!(x_temp, x, A)
+    scale!(x_temp, 1./(A.dx^A.derivative_order))
+    A.t[] += 1 # incrementing the internal time stamp
```
ChrisRackauckas: Time should be updated in `update_coefficients!` directly since many algorithms need to be more careful of this, like when being used in adaptive time stepping algorithms.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -0,0 +1,211 @@
...
+    @boundscheck checkbounds(A, i, j)
+    mid = div(A.stencil_length, 2) + 1
+    bpc = A.stencil_length - mid
+    l = max(1, low(j, mid, bpc))
+    h = min(A.stencil_length, high(j, mid, bpc, A.stencil_length, A.dimension))
+    slen = h - l + 1
+    if abs(i - j) > div(slen, 2)
+        return 0
+    else
+        return A.stencil_coefs[mid + j - i]
```
ChrisRackauckas: Looks like for now this just does the stencil matrix without boundary conditions?

src/DiffEqOperators.jl
```diff
@@ -14,11 +14,13 @@ include("diffeqscalar.jl")
...
 include("derivative_operators/fornberg.jl")
+include("derivative_operators/upwind_operator.jl")
+include("derivative_operators/derivative_irreg_operator.jl")
+include("derivative_operators/derivative_operator.jl")
+include("derivative_operators/abstract_operator_functions.jl")
 include("derivative_operators/boundary_operators.jl")
 
 export DiffEqScalar, DiffEqArrayOperator
-export AbstractDerivativeOperator, DerivativeOperator, UpwindOperator
+export AbstractDerivativeOperator, DerivativeOperator, UpwindOperator, FiniteDifference
```
ChrisRackauckas: Not such a fan of the name FiniteDifference. We should bikeshed that a little bit.

ChrisRackauckas:
I really like the refactor. Thanks for doing that.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.6
 DiffEqBase 1.19.0
 LinearMaps
+DiffEqOperators
```
ChrisRackauckas: You don't have to require yourself.

src/derivative_operators/derivative_operator.jl
```diff
@@ -45,10 +45,10 @@ end
 
 function negate!(arr::T) where T
     if size(arr,2) == 1
-        scale!(arr,-one(eltype(arr)))
+        scale!(arr,-one(eltype(arr[1]))) #fix right neumann bc, eltype(Vector{T}) doesnt work.
```
ChrisRackauckas: Really? `eltype(Vector{T})` returns `T`.

test/runtests.jl
```diff
@@ -1,6 +1,6 @@
 using DiffEqOperators
 using Base.Test
-using SpecialMatrices
+using SpecialMatrices, SpecialFunctions
```
ChrisRackauckas: This needs to go in test/REQUIRE

src/derivative_operators/derivative_operator.jl
fgerick: Sorry, what I meant what doesn't work is this:``eltype(arr)=eltype(Array{Array{T,1},1})=Array{T,1}`` and the ``one`` method fails.

ChrisRackauckas: Got it.

ChrisRackauckas:
Rebase to master to fix that KdV test (well, at least have it report failure without exiting)

shivin9:
@ChrisRackauckas yeah I saw this today morning but didn't get time to look into it. Will do it tomorrow.

src/DiffEqOperators.jl
jlperla: I am very excited about this functionality, and would be happy to help get some tests created.

 Since I see the current `DerivativeOperator` as a special case of this, consider whether the naming should be more like
- `DerivativeOperator` to `RegularDerivativeOperator`
- `FiniteDifference` to `IrregularDerivativeOperator`

The thing I don't understand about the interface is how much of the grid it is keeping around, at it looked to me like it only kept the `dx` rather than the whole `x` grid.  

For example, creating it with `FiniteDifference{Float64}(2,2,dx,length(x),:Dirichlet0,:Dirichlet0,BC=(0.0,0.0))` how do you know the exact boundaries it is valid on?  Those numbers might become very useful when we look at boundary value problems which create affine operators.

When operators need to be composed together, I think it will be very important to be able to do checks on the boundaries/etc?

fgerick: I would suggest to name use the names ``RegularOperator`` and ``IrregularOperator``, I find the ``Derivative`` redundant, as they're both subtypes of ``AbstractDerivativeOperator``. You also don't say ``UpwindDerivativeOperator`` (Although I guess you could do that as well...).

For the second part I'm not sure I understand the problem.

ChrisRackauckas: Since there are other non-FDM operators, it might make sense to call it `AbstractFDMOperator` with `RegularFDMOperator` and `IrregularFDMOperator`.

I don't think the boundary values are required because they can be enclosed.

src/derivative_operators/abstract_operator_functions.jl
ChrisRackauckas: Yeah since that's allocated, might as well get a view from it.

ChrisRackauckas: Well actually, the type would be different than what one might expect... we can discuss that further in an issue though.

ChrisRackauckas:
I think there are still a few things to be done in this domain, but I like these changes and we will be revamping BCs soon ( https://github.com/JuliaDiffEq/DiffEqOperators.jl/issues/53 ) so I'm going to pull this in so that way it gets treatment during the revamp. Thanks @fgerick ! We'll thus continue develop and discussion of the naming changes in https://github.com/JuliaDiffEq/DiffEqOperators.jl/issues/53

Issue #53, opened 2018-03-03
BC Overhaul


ChrisRackauckas:
Long discussions with @dlfivefifty and @jlperla led to a big shift in the way we are doing BCs. The idea is outlined https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/260#issuecomment-370111767 . For DiffEqOperators.jl to change to support this, we need to do the following:

1. The DerivativeOperators should be the non-square `A` interior stencil calculations. All interior, no boundaries. `full`, `sparse`, `banded`, etc. should be defined on the stencil to make it easy to get matrices out.
2. Lazy `R` should be created. That's super simple of course. And full/space/etc. here too.
3. Lazy `Q` should be created for each of the FDM boundary types. So there should be a constructor `FDMExtension(...)` which takes in boundary types and spits out an operator which does this operation. `Q` can be an affine operator. We'll need to define that such that `(Qa,Qb) = full(Q)`. And full/space/etc. here too.
4. In this form, `v_t = L*v` can be written at `v_t = A*Qa*v + A*Qb`.
5. We should have a lazy `L`, dependent on the BCs, which is `L*v = A*Qa*v + A*Qb`. 
6. In this form, `(L1+L2)*v` is well-defined as `(A1+A2)*Qa*v + (A1+A2)*Qb`.
7. We should rename `DerivativeOperator` to `FDMDerivativeOperator` and then `FiniteDifference` to `FDMIrregularDerivativeOperator` (or something like that). Both `<: AbstractDerivativeOperator`. 

Every step of the way should get tests of course.

jlperla:
@stevenzhangdx @FernandoKuwer   Note, even more description for your writeup and test cases.

PR #54, opened 2018-03-03
Operator norm

MSeeker1340:
This is mostly to prepare for the exponential Krylov integrators using _Expokit.jl_, which require the operator's norm for error estimation.

For the derivative operators, the Inf norm can be calculated efficiently using the stencil coefficients (this is also what _Expokit_ requires). The other norms default to calling `Base.norm` on the full matrix.

 _Expokit_ accepts both `DerivativeOperator` and `UpwindOperator` now:

```julia
using DiffEqOperators
using Expokit: expmv
N = 1000
dx = 0.1
dt = 0.1
tol = 1e-5
u = rand(N)

L1 = DerivativeOperator{Float64}(2,4,dx,N,:Dirichlet0,:Dirichlet0)
A1 = full(L1)
@show isapprox(expm(A1*dt)*u, expmv(dt, L1, u); rtol=tol) # true

L2 = UpwindOperator{Float64}(2,4,dx,N,BitVector(N),:Dirichlet0,:Dirichlet0)
A2 = full(L2)
@show isapprox(expm(A2*dt)*u, expmv(dt, L2, u); rtol=tol) # true
```

Note that the calculated norm does not make sense for Dirichlet1 and Neumann1 BC currently because operators with these BC aren't even linear in the first place. This should not be a problem after the BC overhaul #53 is done because the interior section `u[iidxs]` does in fact form a vector space, unlike the full domain `u`.

### DIFF
--- a/src/array_operator.jl
+++ b/src/array_operator.jl
@@ -63,6 +63,7 @@ function Base.expm(L::DiffEqArrayOperator)
 end
 DiffEqBase.has_expm(L::DiffEqArrayOperator) = true
 Base.size(L::DiffEqArrayOperator) = size(L.A)
+Base.norm(L::DiffEqArrayOperator, p::Real=2) = norm(L.A, p)
 DiffEqBase.update_coefficients!(L::DiffEqArrayOperator,t,u) = (L.update_func(L.A,t,u); L.α = L.α(t); nothing)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Nice!

PR #55, opened 2018-03-05
Linear Combinations of Operators

MSeeker1340:
Adds `expm` and `norm` methods for linear combinations of `DiffEqArrayOperator` and derivative operators. Also includes a `normbound` method that computes a upper bound to the norm using triangle inequality. `normbound(A,Inf)` can serve as an efficient estimate for the operator norm in applications where the exact norm is not desired (for example in error estimation in Krylov methods), because it is easy to calculate the Inf norm for derivative operators.

In addition, fixes an error in `norm` for `DiffEqArrayOperator` (it now takes into account the coefficient too) and adds a test set for `DiffEqArrayOperator`.

The efficient `full` method for linear combinations is currently bugged as Julia's type dispatching will always default to the `full` method defined in LinearMaps.jl.

Not sure where to put the functions for the linear combination though, since it handles both `DiffEqArrayOperator` and derivative operators.

### DIFF
--- a/src/array_operator.jl
+++ b/src/array_operator.jl
@@ -52,15 +52,8 @@ Base.issymmetric(L::DiffEqArrayOperator) = L._issymmetric
 Base.ishermitian(L::DiffEqArrayOperator) = L._ishermitian
 Base.isposdef(L::DiffEqArrayOperator) = L._isposdef
 DiffEqBase.is_constant(L::DiffEqArrayOperator) = L.update_func == DEFAULT_UPDATE_FUNC
-function Base.expm(L::DiffEqArrayOperator)
-    tmp = full(L.A) # If not lazy then this is a no-op
-    tmp .*= L.α.coeff
-    out = expm(tmp)
-    if tmp === L.A
-        L.A ./= L.α.coeff # Undo change if not lazy
-    end
-    out
-end
+Base.full(L::DiffEqArrayOperator) = full(L.A) * L.α.coeff
+Base.expm(L::DiffEqArrayOperator) = expm(full(L))
 DiffEqBase.has_expm(L::DiffEqArrayOperator) = true
 Base.size(L::DiffEqArrayOperator) = size(L.A)
 Base.norm(L::DiffEqArrayOperator, p::Real=2) = norm(L.A, p) * abs(L.α.coeff)
--- a/src/array_operator.jl
+++ b/src/array_operator.jl
@@ -63,7 +63,7 @@ function Base.expm(L::DiffEqArrayOperator)
 end
 DiffEqBase.has_expm(L::DiffEqArrayOperator) = true
 Base.size(L::DiffEqArrayOperator) = size(L.A)
-Base.norm(L::DiffEqArrayOperator, p::Real=2) = norm(L.A, p)
+Base.norm(L::DiffEqArrayOperator, p::Real=2) = norm(L.A, p) * abs(L.α.coeff)
 DiffEqBase.update_coefficients!(L::DiffEqArrayOperator,t,u) = (L.update_func(L.A,t,u); L.α = L.α(t); nothing)
 DiffEqBase.update_coefficients(L::DiffEqArrayOperator,t,u)  = (L.update_func(L.A,t,u); L.α = L.α(t); L)
 
--- a/src/array_operator.jl
+++ b/src/array_operator.jl
@@ -52,7 +52,7 @@ Base.issymmetric(L::DiffEqArrayOperator) = L._issymmetric
 Base.ishermitian(L::DiffEqArrayOperator) = L._ishermitian
 Base.isposdef(L::DiffEqArrayOperator) = L._isposdef
 DiffEqBase.is_constant(L::DiffEqArrayOperator) = L.update_func == DEFAULT_UPDATE_FUNC
-Base.full(L::DiffEqArrayOperator) = full(L.A) * L.α.coeff
+Base.full(L::DiffEqArrayOperator) = full(L.A) .* L.α.coeff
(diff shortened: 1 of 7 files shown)


MSeeker1340:
Seems I still haven't covered everything thoroughly...

To calculate caching phi in `NorsettEuler` I need to evaluate
```julia
expA = expm(A*dt)
phi1 = ((expA-I)/A)
```
so there's more work.

src/operator_combination.jl
```diff
@@ -0,0 +1,43 @@
...
+#             @. out += c * full(op)
+#         end
+#     end
+#     return out
+# end
+
+Base.expm(A::LinearCombination) = expm(full(A))
+
+Base.norm(A::IdentityMap{T}, p::Real=2) where T = real(one(T))
+Base.norm(A::LinearCombination, p::Real=2) = norm(full(A), p)
```
ChrisRackauckas: There's a lazy way to do `normest2` which will be required when `A` doesn't fit into memory, but we can handle that later.

MSeeker1340:
I'll drop the WIP flag once things checkout on the OrdinaryDiffEq end.

Also, should we explicitly call `full` at the caching step? i.e. instead of 
```julia
expA = expm(A*dt)
phi1 = (expA - I) / A
```
use
```julia
fullA = full(A)
expA = expm(fullA*dt)
phi1 = (expA - I) / fullA
```

This way we will avoid a lot of potential problems for future caching methods.

ChrisRackauckas:
>Also, should we explicitly call full at the caching step? 

Yes, for the cached ones we know we have to `full` so we might as well do it right before the computations.

MSeeker1340:
All set. I'll do a PR on OrdinaryDiffEq after this is updated.

src/array_operator.jl
```diff
@@ -52,18 +52,11 @@ Base.issymmetric(L::DiffEqArrayOperator) = L._issymmetric
...
-function Base.expm(L::DiffEqArrayOperator)
-    tmp = full(L.A) # If not lazy then this is a no-op
-    tmp .*= L.α.coeff
-    out = expm(tmp)
-    if tmp === L.A
-        L.A ./= L.α.coeff # Undo change if not lazy
-    end
-    out
-end
+Base.full(L::DiffEqArrayOperator) = full(L.A) * L.α.coeff
```
ChrisRackauckas: `.*` this so that way it's compatible with coefficient arrays. That'll come up in space-dependent coefficients and quasi-linear PDEs.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -209,3 +209,11 @@ function Base.sparse(A::AbstractDerivativeOperator{T}) where T
...
     return mat
 end
+
+#=
+    Fallback methods that use the full representation of the operator
+=#
+Base.expm(A::AbstractDerivativeOperator{T}) where T = expm(full(A))
+Base.:/(A::AbstractVecOrMat, B::AbstractDerivativeOperator) = A / full(B)
+Base.:/(A::AbstractDerivativeOperator, B::AbstractVecOrMat) = full(A) / B
+# Base.:\ is also defined
```
ChrisRackauckas: IIRC isn't `/` defined in terms of `\` with a transpose? It may be better to do that one directly. I would use `@which` to dive into the Base code and see how `/` is called.

ChrisRackauckas:
Small changes I missed last time, then this looks good to go.

src/derivative_operators/abstract_operator_functions.jl
MSeeker1340: I commented out this portion and found that 
```julia
using DiffEqOperators
N = 20
A = DerivativeOperator{Float64}(2,2,1.0,N,:Dirichlet0,:Dirichlet0)
mat1 = rand(N,N)

@which A / mat1
no method found for the specified argument types

@which mat1' \ A'
\(x, y) at operators.jl:457
```

The definition is
```julia
\(x,y) = (y'/x')'
```

So Julia does in fact define `\` in terms of '/' for generic types. However for regular matrices the relationship is inverted:
```julia
# In linalg\generic.jl:820
(/)(A::AbstractVecOrMat, B::AbstractVecOrMat) = (B' \ A')'
```

Do you think I should define '/' using this convention, or just go with the generic one?

ChrisRackauckas: I think you might need both. The transpose operator will be hard to do in general, and right now I think it's a no-op which only holds for some BCs and since we have the operator as `n x n`. Transpose on the dense matrix will always work though.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -209,3 +209,16 @@ function Base.sparse(A::AbstractDerivativeOperator{T}) where T
...
+    Fallback methods that use the full representation of the operator
+
+    As with the convention for regular matrices, right division is defined in 
+    terms of left division. (The result may not be correct for some BC as the 
+    transpose of a derivative operator is currently defined to be a no-op)
+=#
+Base.expm(A::AbstractDerivativeOperator{T}) where T = expm(full(A))
+Base.:\(A::AbstractVecOrMat, B::AbstractDerivativeOperator) = A \ full(B)
+Base.:\(A::AbstractDerivativeOperator, B::AbstractVecOrMat) = full(A) \ B
+Base.:/(A::AbstractVecOrMat, B::AbstractDerivativeOperator) = (B' \ A')'
```
ChrisRackauckas: These transposes might be incorrect because they are just implemented as a no-op. Sorry if I was unclear before. I think you should just `full` and `/`, or does the generic version but `full` before transpose.

PR #61, opened 2018-04-13
Column stacking form in the overview

MSeeker1340:
@jlperla We discussed this Monday about the form of the solution. I have changed the overview document to use the column stacking form. This shouldn't affect the operators in any way.

BTW, is the overview document intended as an internal sketch or is it meant to be presented to the users? If it's the latter case, I think it's better to restructure the document to first discuss the univariate case and then include a section for the multivariate generalization (which doesn't really add anything to our approach to the BC). This should make the discussions much easier to follow.

### DIFF
--- a/docs/linear_operators_overview.tex
+++ b/docs/linear_operators_overview.tex
@@ -46,20 +46,19 @@
 		Use $i=1,... I_j$ as the grid in the interior. Moreover, define $I_{j,L}$ and $I_{j,H}$ as the non interior grid points. Therefore, $\hat{I}_j = I_j + I_{j,L} + I_{j,H}$. % Again, I think you mean I_j's in above cases.
 		 Notice that the solution to the discretized function is $u \in R^{\hat{I}}$ if stationary.
 		
-		We construct the discretized $\hat{u}$ as a block matrix of size $\hat{I}\times n$:
+		We construct the discretized $\hat{u}$ by column stacking the discretized $\hat{u}_{x^j}$ on each grid $j \in \{1,...,n\}$:
 		\begin{equation}
 		\hat{u} = \begin{bmatrix}
-		\hat{u}_{x^1} & \mathbf{0}_1       & ...    & \mathbf{0}_1\\
-		\mathbf{0}_2       & \hat{u}_{x^2} & ...    & \mathbf{0}_2\\
-		\vdots  & \vdots  & \ddots & \vdots \\
-		\mathbf{0}_n       & \mathbf{0}_n       & ...    & \hat{u}_{x^n}\\
+		\hat{u}_{x^1}\\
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>BTW, is the overview document intended as an internal sketch or is it meant to be presented to the users?

Internal sketch. I thought at first that it would directly lead to code but it seems to need more discussion. We should probably move this out to a separate repo since TeX documents and PDFs will blow the repo size, so that branch can't/shouldn't ever be merged. I created this repo awhile back and it should probably be used for this:

https://github.com/JuliaDiffEq/DiffEqDevMaterials

In fact, I'm not entirely sure about the best way to start organizing this, but work with people like @YingboMa have lead to Mathematica notebook derivations and other internal documents that we need to start being more careful about storing for future use.

ChrisRackauckas:
One other thing I am interested in looking into is if there's a better way to do this document writing collaboratively in a way that generates the output. @jlperla and I discussed using Markdown with https://chrome.google.com/webstore/detail/github-with-mathjax/ioemnmodlmafdkllaclgeombjnmnbima, but that's limited in its transferability. I do have a paid Overleaf account so we could use that to share documents from Github repos. I'd have to investigate how to set the Git integration up.

jlperla:
The document is definitely a work in progress, and it is in an especially rough state at the moment.  @FernandoKuwer and @stevenzhangdx  are planning to radically simplify to basic description (i.e., remove the current multivariate case, which was premature) and make it more generically by becoming simpler.  They will be able to get back to it after the exams and assignments.  No doubt, however, that the stacked column would be the way to go.

We tried the markdown instead of latex, but it didn't work very well.  The equation references/macros/etc. were a pain and my hoped for pandoc output just didn't work well.  I think that the markdown docs are a good approach for simple tutorials, but not for a bunch of interconnected equations.  But maybe there is a way that works that I do not know?  As this should be the mathematical design document that goes through all the algebra, it also may be the only one that strictly required latex.

Keep in mind that right now the PDF files and all intermediates are in the .`gitignore` , so I don't see how it would clutter the repo.  I hear you about mathematica notebooks, etc. which I frequently require for these sorts of projects.  If you want us to move the document into https://github.com/JuliaDiffEq/DiffEqDevMaterials instead, just give us the word.  Then we can leave the univariate_tests for what we had originally intended: the example code for testing.  For example,  we think https://github.com/JuliaDiffEq/DiffEqOperators.jl/blob/univariate_tests/docs/operator_examples/simple_stationary_HJBE_reflecting.jl is the correct implementation using the notation and the boundary values... to be checked.

ChrisRackauckas:
I'll merge but let's move this out of the repo and close the branch. It's not really an issue yet, but sooner rather than later we should start putting figures into any writing which is where the problem begins.

PR #76, opened 2018-08-28
WIP: Generic Derivative Operators

MSeeker1340:
The aim is to implement the generic `L*QB` derivative operators as described by the dev document of https://github.com/JuliaDiffEq/PDERoadmap. I'll leave the current operators unchanged and prep this as an developmental branch of sorts to experiment on.

What we need:

- [x] Stencil operator (extended domain -> interior) types which use the Fornberg algorithm to generate stencil coefficients and carry out stencil convolution operations.

- [ ] ~~Boundary extrapolation operators (interior -> extended domain). Most likely will use LazyArrays.jl to make these more efficient memory-wise.~~

- [x] `GenericDerivativeOperator` which is basically one stencil operator (or a combination of them) and a boundary extrapolation operator. 

Currently I've finished the first part and modeled the `UniformDerivativeStencil`, `IrregularDerivativeStencil` and `UniformUpwindStencil` types on the current `DerivativeOperator`, `FiniteDifference` and `UpwindOperator` (I can add a `IrregularUpwindStencil` later). Basically I just removed all parts related to the boundary conditions, along with some minor changes (e.g. storing the entire size in `L.dimension` as they are now rectangular).

For the upwind stencils, I've chosen the option 2 route of https://github.com/JuliaDiffEq/PDERoadmap/pull/34#issuecomment-415824216, i.e. have the coefficients be handled at the stencil level so that they can function alone and be freely composed. One change I've made is that instead of embedding the coefficients in the stencil types, I instead utilized the scaled operator type that is generated naturally by left multiplying a scalar and overloaded the multiplication method for it. I think this can make the expression tree easier to analyze and the `UniformUpwindStencil` type interface simpler.

I'm not 100% sure about the correctness of the stencil operation though. I can't really decipher the semantics of `UpwindOperator`, in particular what's the deal with `stencil_rem`. I've checked that the simple first-order upwind derivative is ok but somebody with more experience on the subject should probably take a closer look at this.

### DIFF
--- a/src/derivative_operators/derivative_stencil.jl
+++ b/src/derivative_operators/derivative_stencil.jl
@@ -15,6 +15,17 @@ struct UniformDerivativeStencil{T,S<:SVector} <: AbstractDiffEqLinearOperator{T}
     UniformDerivativeStencil(xgrid::AbstractRange{T},dorder,aorder) where {T} =
         UniformDerivativeStencil(dorder,aorder,step(xgrid),length(xgrid))
 end
+function mul!(y::AbstractVector{T}, L::UniformDerivativeStencil{T,S}, x::AbstractVector{T}) where {T,S}
+    coeffs = L.stencil_coefs
+    Threads.@threads for i in 1:length(y)
+        ytemp = zero(T)
+        @inbounds for idx in 1:L.stencil_length
+            ytemp += coeffs[idx] * x[i + idx - 1]
+        end
+        y[i] = ytemp
+    end
+    return y
+end
 
 struct IrregularDerivativeStencil{T,S<:SVector} <: AbstractDiffEqLinearOperator{T}
     derivative_order    :: Int
@@ -32,7 +43,22 @@ struct IrregularDerivativeStencil{T,S<:SVector} <: AbstractDiffEqLinearOperator{
         new{T,eltype(stencil_coefs)}(dorder,aorder,dim,stencil_length,stencil_coefs)
     end
 end
+function mul!(y::AbstractVector{T}, L::IrregularDerivativeStencil{T,S}, x::AbstractVector{T}) where {T,S}
+    coeffs = L.stencil_coefs
+    Threads.@threads for i in 1:length(y)
+        ytemp = zero(T)
+        @inbounds for idx in 1:L.stencil_length
+            ytemp += coeffs[i][idx] * x[i + idx - 1]
+        end
+        y[i] = ytemp
+    end
+    return y
+end
 
 DerivativeStencil = Union{UniformDerivativeStencil, IrregularDerivativeStencil}
 size(L::DerivativeStencil) = L.dimension
 size(L::DerivativeStencil, i::Int) = i <= 2 ? L.dimension[i] : 1
+function *(L::DerivativeStencil, x::AbstractVector)
+    y = zeros(promote_type(eltype(L), eltype(x)), size(L, 1))
+    mul!(y, L, x)
+end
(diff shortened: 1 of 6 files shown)


MSeeker1340:
I tried writing standalone Q operators using LazyArrays.jl but can't come up with a good way to do this. As @ChrisRackauckas pointed out the `mul!` method would be difficult to write, as we would need to break the lazy concatenation interface of LazyArrays.jl and manually update the field arrays of a `VCat`. I thought about using caches in the Q operators to avoid having to allocate for the boundary points, but then the resulting `VCat` will point back to the caches in Q and that leads to all sorts of reference problems.

So instead, I implemented the Q operators as implicit components of the `GenericDerivativeOperator` type. Though this is similar to how `DerivativeOperator` is currently doing, the main difference is that `GenericDerivativeOperator` is not limited to one stencil (as specified by a `dorder` and `aorder`) but can be a combination of stencils (or just a general linear operator). The `mul!` method in this case is clearly divided in two stages: first the implicit Q operator is applied to generate an `xbar` as a `Vcat` of boundary points (pre-allocated as caches) and `x`, then `L` is applied to xbar. Since `xbar` is now a temporary object, the reference issue in the standalone Q version no longer exists.

For an example of how the low-level interface looks like, see the test script in `test/generic_derivative_operator.jl`

I believe this is a good way to implement the generic `L*QB` operators, i.e. having standalone stencil operators `L` and implicit `QB`. Not sure if this can replace the current derivative operators though;  I think performance analysis is needed.

Of course, the handling of boundary conditions/Q is still very crude at the moment. In particular, the current implementation of `applyQ!` and `constructQ` packs everything in a giant conditional, which could be improved e.g. by encoding the BCs as types instead of symbols so they can be dispatched on.

P.S. As for affine Q, the `applyQ!` interface does not need much change because we just need to fill out the boundary points and it does not matter if the action is linear or affine. On the other hand `constructQ` would need drastic changes because we cannot really represent an affine operator as a matrix (actually, the whole `convert(AbstractMatrix,A)` interface would need to change for affine `A`s.

ChrisRackauckas:
>I'm not 100% sure about the correctness of the stencil operation though. I can't really decipher the semantics of UpwindOperator, in particular what's the deal with stencil_rem. I've checked that the simple first-order upwind derivative is ok but somebody with more experience on the subject should probably take a closer look at this.

I think this handling was required for order > 2 stencils. Essentially, when order > 2 then you get more than one point to the left/right. To handle the BCs is a little tricky at that point. What we did was 1-sided differences near the BCs. However, now I know that's less stable, so what it should be is instead a non-symmetric difference of the right order which uses exactly to the boundary point, and the boundary point is then determined by the ghost point value. Now that it's all formalized this is clear how to do, but it wasn't back then!

>Of course, the handling of boundary conditions/Q is still very crude at the moment. In particular, the current implementation of applyQ! and constructQ packs everything in a giant conditional, which could be improved e.g. by encoding the BCs as types instead of symbols so they can be dispatched on.

Yeah.

src/derivative_operators/derivative_stencil.jl
```diff
@@ -0,0 +1,205 @@
...
+        y[i] = ytemp
+    end
+    return y
+end
+function convert(::Type{AbstractMatrix}, L::UniformDerivativeStencil)
+    coeffs = L.stencil_coefs
+    mat = spzeros(eltype(L), size(L,1), size(L,2))
+    for i in 1:size(L,1)
+        for idx in 1:L.stencil_length
+            mat[i, i + idx - 1] = coeffs[idx]
```
ChrisRackauckas: this is a very slow way of doing this, but we can ignore that for now. It would be better to try and make a BandedMatrix anyways.

PR #77, opened 2018-09-11
Taking out of domain points into account

fgerick:
I started to implement correct boundary conditions for higher order schemes when the stencils reaches out of the domain. I included some tests, but more are needed. I also updated the irregular grid operator a bit. Something is wrong in the convolution or the stencil generation (works fine for uniform grid), I would be happy if someone could look into this as well as.

- [x]  ```DerivativeOperator``` higher order approximation boundary conditions ```:None```, ```:Dirichlet```, ```Dirichlet0```.
- [ ]  ```DerivativeOperator``` higher order approximation boundary conditions ```:Neumann```, ```:Neumann0```, ```Robin```.
- [ ]  ```FiniteDifference``` verification of correctness.
- [ ] Translation into new BC formulation.


ChrisRackauckas:
For more discussion on the direction we are going with the BCs and the handling of this, you should probably join in at https://github.com/JuliaDiffEq/PDERoadmap

PR #82, opened 2018-11-01
Make MatrixFreeOperator handle parameters and time-dependent operators

YingboMa:
If the user writes `MatrixFreeOperator(f)` or `MatrixFreeOperator(f, (p,))` we assume it is constant and `f` must be in the form of `(du, u, p) -> ...`, if the user writes `MatrixFreeOperator(f, (p, t))` we assume it is time-dependent and `f` must be in the form of `(du, u, p, t) -> ...`.

See https://github.com/JuliaDiffEq/DiffEqOperators.jl/blob/f5248d71dbcf93302a17057282cce7f29f8b7483/test/matrixfree.jl#L4 for how it looks.


chiyahn:
Awesome. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/547 should be resolved on the PR is merged.

PR #94, opened 2019-05-28
WIP Overload of Array/SparseMatrixCSC/BandedMatrix

ajozefiak:
Work in progress of the overloading of Array, SparseMatrixCSC, and BandedMatrix. First pass of the implementation for the uniform grid operators: DerivativeOperator and UpwindOperator.


ChrisRackauckas:
You're going to need to redo the PR from a new branch and new commit because the DS_STORE binary is in the git history of this commit.

src/DiffEqOperators.jl
```diff
@@ -2,9 +2,11 @@ module DiffEqOperators
 
 import Base: +, -, *, /, \, size, getindex, setindex!, Matrix, convert
 using DiffEqBase, StaticArrays, LinearAlgebra
-import LinearAlgebra: mul!, ldiv!, lmul!, rmul!, axpy!, opnorm, factorize, I
+import LinearAlgebra: mul!, ldiv!, lmul!, rmul!, axpy!, opnorm, factorize, I, Array
+import SparseArrays: SparseMatrixCSC
```
ChrisRackauckas: It's not needed to import the names here.

src/derivative_operators/upwind_operator.jl
```diff
@@ -235,3 +235,43 @@ function LinearAlgebra.opnorm(A::UpwindOperator{T,S,LBC,RBC}, p::Real=2) where {
         opnorm(convert(Array,A), p)
     end
 end
+
+function LinearAlgebra.Array(A::UpwindOperator{T}) where T
```
ChrisRackauckas: These upwind operators need to flip the stencils depending on the current direction. We can ignore this for now since they may change soon.

ChrisRackauckas:
Generally looks good. Needs to handle the boundaries at the end though. When the stencil uses more than 1 value on each side, the non-centered discretizations start being used.

src/derivative_operators/upwind_operator.jl
ajozefiak: @ChrisRackauckas it is not sufficient to check A.directions[i] in order to determine the direction of the ith point of the discretization?

ajozefiak: @ChrisRackauckas it is not sufficient to check A.directions[][i] for which stencil to use for the ith gird point?

PR #95, opened 2019-05-28
created new struct InteriorOperator which acts on the vector and appends ghost points also

ChrisRackauckas:


src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -128,6 +128,13 @@ function LinearAlgebra.mul!(x_temp::AbstractVector{T}, A::Union{DerivativeOperat
     A.t[] += 1 # incrementing the internal time stamp
 end
 
+
+function LinearAlgebra.mul!(x_temp::AbstractVector{T}, A::InteriorOperator{T}, x::AbstractVector{T}) where T<:Real
+    convolve_interior!(x_temp, x, A)
```
ChrisRackauckas: This also needs to handle the left/right because the boundary stencils have to do the non-centering when they are too large, so the edges needed to be handled directly as well. They should probably just be called CenteredDifferenceOperator (and DerivativeOperator should probably go away when these are ready).

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -184,6 +191,25 @@ function Base.convert(::Type{Array}, A::AbstractDerivativeOperator{T}, N::Int=A.
 end
 
 
+function Base.convert(::Type{Array}, A::InteriorOperator{T}, N::Int=A.dimension) where T
```
ChrisRackauckas: In the interior, you'll want to grab the contiguous `.x` portion out of the lazy array so that way you can avoid the slower indexing. It might be a good idea to make and test a lazy array first.

src/derivative_operators/abstract_operator_functions.jl
shivin9: Wait a minute. Didn't we plan that we are separating the boundary and interior operations and therefore are using two different operators, `Q` and `L`?

PR #96, opened 2019-05-30
WIP Overload array constructors

ajozefiak:
A second pass of the implementation of array constructor overloads, for the types DerivativeOperator and FiniteDifference (the central difference implementations). Resolves handling of stencils near boundaries.

### DIFF
--- a/src/derivative_operators/derivative_irreg_operator.jl
+++ b/src/derivative_operators/derivative_irreg_operator.jl
@@ -408,17 +408,17 @@ function LinearAlgebra.Array(A::FiniteDifference{T}) where T
     extended_x = one(T).*[-A.dx[1]; zero(T); cumsum([A.dx; A.dx[N-1]])]
 
     # Apply lower stencils
-    for i in 1:stl_2-1
+    for i in 1:stl_2
         L[i,1:stl] = calculate_weights(A.derivative_order, extended_x[i+1], extended_x[1:stl])
     end
 
     # Apply inner stencils
-    for i in max(stl_2,1):N-stl_2+1
-        L[i,i-stl_2+1:i+stl_2+1] = calculate_weights(A.derivative_order, extended_x[i+1], extended_x[i+1-stl_2:i+1+stl_2])
+    for i in max(stl_2,1)+1:N-stl_2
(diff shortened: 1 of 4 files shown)


src/derivative_operators/derivative_irreg_operator.jl
```diff
@@ -399,3 +399,75 @@ end
...
+    extended_x = one(T).*[-A.dx[1]; zero(T); cumsum([A.dx; A.dx[N-1]])]
+
+    # Apply lower stencils
+    for i in 1:stl_2-1
+        L[i,1:stl] = calculate_weights(A.derivative_order, extended_x[i+1], extended_x[1:stl])
+    end
+
+    # Apply inner stencils
+    for i in max(stl_2,1):N-stl_2+1
+        L[i,i-stl_2+1:i+stl_2+1] = calculate_weights(A.derivative_order, extended_x[i+1], extended_x[i+1-stl_2:i+1+stl_2])
```
ChrisRackauckas: you shouldn't need to re-calculate the weights. These are calculated once and stored in the operators.

ajozefiak: I see that loop. My concern is that initialize_left_boundary!() and initialize_right_boundary!() for the Dirichlet, Dirichlet0, Neumann0, and Periodic boundary conditions do not create any stencilling for hte boundary points. 

I could change the implementation to use the boundary stencils computed in the Neumann, Robin, and None boundary types, however, I am unsure as how to deal with the aforementioned boundary types.

ChrisRackauckas: the stencil should end up independent of the boundary, so code for what it will be and not what it currently is for now.

ajozefiak:
@ChrisRackauckas @shivin9 I've pushed some more commits assuming that the various operators are implemented correctly/fully. I've assumed for the UpwindOperator that the boundary stencils correspond to the upwind direction which does not fit within the grid near a particular boundary. So for instance we can apply the interior forward-difference stencil without issue near the lower boundary, but the backward-difference stencil would exceed the boundary, so the correct backward-difference stencil at a point near the lower boundary is assumed to be stored in A.low_stencil_coefs.

PR #99, opened 2019-06-08
RobinBC

ajozefiak:
This branch was forked from Shivin's branch L_operator. Here I am implementing RobinBC and RobinBCExtended. There is nothing else new implemented, and most of this may be wrong. However, I think it would be good to get feedback on the interface of RobinBC and RobinBCExtended before proceeding. 

One thing to note is  that Array(Q::RobinBC) is ill defined without specifying the dimension either in the overload or in the struct RobinBC. The latter approach may come at a cost of reducing the generality of Q by already specifying its dimension. However, we could include a dimension field in RobinBC and implement Q*u such that the dimension of Q is ignored, and u is simply copied over. Whereas the overload of Array would be the only time we really make use of the dimension field in RobinBC.


ChrisRackauckas:
>One thing to note is that Array(Q::RobinBC) is ill defined without specifying the dimension either in the overload or in the struct RobinBC. The latter approach may come at a cost of reducing the generality of Q by already specifying its dimension. However, we could include a dimension field in RobinBC and implement Q*u such that the dimension of Q is ignored, and u is simply copied over. Whereas the overload of Array would be the only time we really make use of the dimension field in RobinBC.

The overload of Array is the time to do this. Notice that is not uncommon in Julia BTW. `Array(undef,N,N)`, `Array(I,N,N)`. Now `Array(Q,N,N)`.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -184,6 +191,25 @@ function Base.convert(::Type{Array}, A::AbstractDerivativeOperator{T}, N::Int=A.
...
+    gpc = A.ghost_point_count
+    mat = zeros(T, (N, N + 2*gpc))
+    v = zeros(T, N + 2*gpc)
+    for i = 1 + gpc:N+gpc
+        v[i] = one(T)
+        #=
+            calculating the effect on a unit vector to get the matrix of transformation
+            to get the vector in the new vector space.
+        =#
+        mul!(view(mat, i-gpc, :), A, v)
```
ChrisRackauckas: We should really do this from the stencil directly, but this will work.

src/derivative_operators/dirichlet_bc_extended.jl
```diff
@@ -0,0 +1,76 @@
+struct DirichletBC{T<:Real} <: AbstractDerivativeOperator{T}
```
ChrisRackauckas: There's no reason to keep the Dirichlet one around when Robin is done.

src/derivative_operators/robin_bc_extended.jl
```diff
@@ -0,0 +1,78 @@
...
+    l::T
+    r::T
+    u::T2
+
+    function RobinBCExtended(u::T2, al::T, bl::T, cl::T,
+                                    dx_l::T, ar::T, br::T, cr::T, dx_r::T) where
+                                    {T,T2<:AbstractVector{T}}
+        u                    = u
+        l = (cl - bl*u[1]/dx_l)*(1/(al-bl/dx_l))
+        r = (cr + br*u[length(u)]/dx_r)*(1/(ar+br/dx_r))
```
ChrisRackauckas: `u[end]`

src/derivative_operators/robin_bc_extended.jl
```diff
@@ -0,0 +1,78 @@
...
+struct RobinBCExtended{T,T2<:AbstractVector{T}}
+    l::T
+    r::T
+    u::T2
+
+    function RobinBCExtended(u::T2, al::T, bl::T, cl::T,
+                                    dx_l::T, ar::T, br::T, cr::T, dx_r::T) where
+                                    {T,T2<:AbstractVector{T}}
+        u                    = u
+        l = (cl - bl*u[1]/dx_l)*(1/(al-bl/dx_l))
```
ChrisRackauckas: Notice how this is just encoding `u' = (u[1] - u[0]) / dx`. The next step will be to generalize this to higher stencils.

ChrisRackauckas:
Make this commit its own PR, do the change for `u[end]`, add tests, and we can merge. There's no reason to get these two PRs mixed up.

PR #103, opened 2019-06-10
Concretization of L operators

ajozefiak:
Thus far, overloads of Array, SparseMatricCSC, and BandedMatrix are implemented for DerivativeOperator.


ajozefiak:
Changed BandedMatrix concretization to use a zero-valued array as opposed to an uninitialized array. Also, incorrect bound in right boundary convolution was being handled correctly by @inbounds, however, was introducing machine error.

PR #106, opened 2019-06-12
[WIP]: Stencil fix

shivin9:
Before fix:
``` julia
julia> A.low_boundary_coefs*6
1-element Array{StaticArrays.SArray{Tuple{5},Float64,1,5},1}:
 [5.5, -10.0, 3.0, 2.0, -0.5]

julia> A.stencil_coefs
5-element StaticArrays.SArray{Tuple{5},Float64,1,5}:
 -0.08333333333333333
  1.3333333333333333 
 -2.5                
  1.3333333333333335 
 -0.08333333333333333
```

After fix:
``` julia
julia> A.low_boundary_coefs*6
2-element Array{StaticArrays.SArray{Tuple{6},Float64,1,6},1}:
 [22.5, -77.0, 107.0, -78.0, 30.5, -5.0]
 [5.0, -7.5, -2.0, 7.0, -3.0, 0.5]      

julia> A.stencil_coefs
5-element StaticArrays.SArray{Tuple{5},Float64,1,5}:
 -0.08333333333333333
  1.3333333333333333 
 -2.5                
  1.3333333333333335 
 -0.08333333333333333
```

For reference compare with https://en.wikipedia.org/wiki/Finite_difference_coefficient


ChrisRackauckas:
This has an extra stencil at the boundaries. Remember, because it's a N x (N+2) operator, the last stencil on the sides are the `[b,0,x,x,x,x]` stencils, not the `[0,x,x,x,x,x]` stencils, since you're never solving for the derivative at the boundary point.

ChrisRackauckas:
Tests are failing.

shivin9:
Its a bounds error problem. I'll post a fix in some time.

src/derivative_operators/convolutions.jl
```diff
@@ -24,65 +25,66 @@ end
...
+        @inbounds for idx in A.boundary_stencil_length:-1:1
             xtempi += coeffs[i][end-idx] * x[end-idx]
         end
-        x_temp[end-A.boundary_length+i] = xtempi
+        x_temp[end-A.boundary_point_count+i-1] = xtempi
     end
 end
 
 ###########################################
+# NOT NEEDED ANYMORE
```
ChrisRackauckas: These are definitely needed, otherwise the operator will be slower due to the indexing fallback.

shivin9: ok.

test/derivative_operators_interface.jl
```diff
@@ -69,7 +69,8 @@ end
     y = convert(Array{BigFloat, 1}, y)
 
     A = DerivativeOperator{BigFloat}(d_order,approx_order,one(BigFloat),N)
-    @test_broken mat = convert(Array, A, N)
+    # mat = convert_by_multiplication(Array,A,N) # badly broken!
```
ChrisRackauckas: why?

test/derivative_operators_interface.jl
```diff
@@ -69,7 +69,8 @@ end
...
+    # @test mat = convert(Array, A, N)
```
ChrisRackauckas: never comment out tests

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -197,3 +233,32 @@ function *(A::AbstractDerivativeOperator,B::AbstractDerivativeOperator)
     #       the derivative of order A.dorder + B.dorder of
     #       approximation_order = min(approx_A, approx_B)
 end
+
+
+function convert(::Type{Array}, A::AbstractDerivativeOperator{T}, N::Int=A.dimension) where T
```
ChrisRackauckas: why have two of this function? Please see the concretization.jl file. We don't need two.

PR #107, opened 2019-06-13
Added test for correctness and equivalance of DerivativeOperator and FiniteDifference - Added more BCs and extended Robin

xtalax:
Compares the aforementioned operators' outputs with an analytically known result, for order up to 6, approximation order up to 8, including operation on a complex function.
Will likely need updating after the new operator interface is finalized.


xtalax:
Got carried away and wrote BC operators to any order for Periodic, Dirichlet,  Neumann, Robin and 'Generalized robin'. 

See the comment above RobinBC for an idea as to how this works.

The original interface is still supported, so the test in robin.jl should still pass.
Also added supertypes for AffineBC and AbstractBC, to ease defining functions that act on these.

All affine BCs have the same fields, allowing for easy multiple dispatch on BCs of this type.

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,160 @@
...
+
+  the variables in correspond to al*u(0) + bl*u'(0) = cl
+
+  Write vector b̄₁ as a vertical concatanation with b0 and the rest of the elements of b̄ ₁, denoted b̄`₁, the same with ū into u0 and ū`. b̄`₁ = b̄`_2 = fill(β/Δx, length(stencil)-1)
+  Pull out the product of u0 and b0 from the dot product. The stencil used to approximate u` is denoted s. b0 = α+(β/Δx)*s[1]
+  Rearrange terms to find a general formula for u0:= -b̄`₁̇⋅ū`/b0 + γ/b0, which is dependent on ū` the robin coefficients and Δx.
+  The non identity part of Qa is qa:= -b`₁/b0 = -β.*s[2:end]/(α+β*s[1]/Δx). The constant part is Qb = γ/(α+β*s[1]/Δx)
+  do the same at the other boundary (amounts to a flip of s[2:end], with the other set of boundary coeffs)
+"""
+
```
ChrisRackauckas: I think you need to remove this space in order to have the docstring usable with `?`

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,160 @@
...
+        b_l = cl/(al+bl*s[1]/dx_l)
+        b_r = cr/(ar+br*s[1]/dx_r)
+
+        return new{T, typeof(a_l)}(a_l, b_l, a_r, b_r)
+    end
+end
+
+"""
+Implements a generalization of the Robin boundary condition, where α is a vector of coefficients.
+Represents a condition of the form α[1] + α[2]u[0] + α[3]u'[0] + α[4]u''[0]+... = 0
```
ChrisRackauckas: wow, nice generalization!

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,160 @@
+abstract type AbstractBC{T} end
+
+# robin, general, and in general neumann BCs are all affine opeartors, meaning that they take the form Qx = Qax + Qb. neumann0 is not however; is a specialization needed?
+abstract type AffineBC{T,V} <: AbstractBC{T} end
+
+struct DirichletBC{T} <: AbstractBC{T}
```
ChrisRackauckas: Dirichlet is also affine if `l` and `r` aren't 0.

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,160 @@
...
+    l::T
+    r::T
+end
+
+struct NeumannBC{T, V<:AbstractArray{T}} <: AffineBC{T,V}
```
ChrisRackauckas: I think a simpler implementation might be to have `DirichletBC` and `NeumannBC` be simplified constructors for `RobinBC`, but just construct a `RobinBC` struct?

test/runtests.jl
```diff
@@ -5,10 +5,9 @@ import Base: isapprox
 @time @safetestset "Robin Boundary Condition Operators" begin include("robin.jl") end
 @time @safetestset "JacVec Operators Interface" begin include("jacvec_operators.jl") end
 @time @safetestset "Composite Operators Interface" begin include("composite_operators_interface.jl") end
-
 @time @safetestset "Derivative Operators Interface" begin include("derivative_operators_interface.jl") end
+#@time @safetestset "Validate and Compare Generic Operators" begin include("generic_operator_validation.jl") end
```
ChrisRackauckas: why comment it?

ChrisRackauckas:
I really like this PR. This is exactly where we needed to go. But we should condense a bit and get Dirichlet and Neumann into RobinBC so that way it's easier to modify/maintain in the near future (for when we want to add directionality to the end derivatives).

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,160 @@
...
+    b_r::T
+    function GeneralBC{T,V}(αl::AbstractArray{T}, αr::AbstractArray{T}, dx::AbstractArray{T}, order::T = one(T)) where {T,V<:AbstractArray}
+        dx_l, dx_r = dx
+        nl = length(αl)
+        nr = length(αr)
+        S_l = zeros(T, (nl-2, order+nl-2))
+        S_r = zeros(T, (nr-2, order+nr-2))
+
+        for i in 1:(nl-2)
+            S_l[i,:] = [transpose(calculate_weights(i, zero(T), zero(T):convert(T, (order+i))) transpose(zeros(T, nl-2-i))] #am unsure if the length of the dummy_x is correct here
```
ChrisRackauckas: @shivin9 take a look at this here.

test/runtests.jl
xtalax: Because it wasn't passing, I wasn't sure if the interface to DerivativeOperator was going to change after #106  is merged, so I didn't check that too thoroughly - I'll fix that after the merge

ChrisRackauckas: Okay. Normally, just keep the tests in an mark them `@test_broken` so they continue to be tracked.

src/derivative_operators/BC_operators.jl
xtalax: Thanks! I'm well proud of that, it was super satisfying to see all the terms condense down like that.

src/derivative_operators/BC_operators.jl
xtalax: Done in next commit

src/derivative_operators/BC_operators.jl
xtalax: fixed comment

xtalax:
@shivin9 
For clarification, order here represents the approximation order, and i is the derivative order (dorder is always 1 for robin). 
I was thinking that the user supplies it upon construction of the BC operator, or perhaps aorder is taken to be the same as the order of approximation used by the DerivativeOperator, and the BC operator can be constructed from a Derivative Operator to achieve this, or they can both be constructed together. Since a derivative operator is intrinsically tied to an axis, it might be useful for the future to have a DerivativeOperator carry around the axis that it is defined on rather than just dimension and dx

xtalax:
Since a derivative operator (and for now BC operators) is intrinsically tied to an axis, it might be useful for the future to have a DerivativeOperator carry around the axis that it is defined on rather than just dimension and dx, unless there is a good reason why this is not already the case. The BCs will need to be constructed differently also to account for non uniform grids, since their stencils can now extend further than the point adjacent to the boundary, dx_l/r are vector valued in this case.

ChrisRackauckas:
>Since a derivative operator (and for now BC operators) is intrinsically tied to an axis, it might be useful for the future to have a DerivativeOperator carry around the axis that it is defined on rather than just dimension and dx, unless there is a good reason why this is not already the case.

One step at a time. Multidimensional will go next, but we should get one dimension solid first. Likely we will want to have axis in the type information for dispatch purposes.

ChrisRackauckas:
>I was thinking that the user supplies it upon construction of the BC operator, or perhaps aorder is taken to be the same as the order of approximation used by the DerivativeOperator, and the BC operator can be constructed from a Derivative Operator to achieve this, or they can both be constructed together.

Optional and defaulting to 1 is fine. We can do fancy order matching in the next stage if it seems necessary, but many times we might want user input (the order doesn't have to match the interior stencil to get the higher order solution)

test/generic_operator_validation.jl
```diff
@@ -0,0 +1,45 @@
...
+y_ = y[2:(end-1)]
+
+
+
+for dor in 1:6, aor in 1:8
+
+    D1 = DerivativeOperator{Float64}(dor,aor,dx[1],length(x))
+    D2 = DiffEqOperators.FiniteDifference{Float64}(dor,aor,dx,length(x))
+    D = (D1,D2)
+    @test_broken convert(Array, D1) ≈ convert(Array, D2)
```
ChrisRackauckas: while testing operators against operators is good, at some point we need to hard code some solutions to check against known analytical cases.

ChrisRackauckas:
@xtalax in the future, it will be easier if you work on a branch in your fork. I notice this is on master on your fork, which is fine, but can make it a little more difficult to track/rebase.

xtalax:
Ok. This is my first time contributing to an open source project with other contributors, so I don't have any idea about best practices. Can you recommend a guide? Or is it just something that comes with experience?

ChrisRackauckas:
It just comes with experience.

ChrisRackauckas:
@xtalax I took a look through it and it looks nice. I fixed a few dispatch issues and made the operator building more duck typed. However, in order to merge we should make sure this at least passes @ajozefiak 's Robin BC tests. Currently it fails those which is an issue, pointing to maybe a problem with the stencil that is being generated. This kind of goes back to what I mentioned earlier up there: we cannot keep writing tests that test generated operators against generated operators because that won't catch an error in the operator generation process. We need to test against some concrete cases, so let's at least get these working.

ajozefiak:
@xtalax I have to double check my derivations again for the tests and the a_l and b_l's (as you are using them). At the moment it looks to me like my tests are correct. 

I also have the suggestion that the notation of `al` and `a_l` (and etc) should be changed to distinguish more clearly from `a` the `b`'s of `au_0 + bu'_0 = c`  and the `a_l` and `b_l` that correspond to the linear and affine components of the operator we are computing. Additionally, down the road when we will generalize to the non-uniform grid case, the `dx`'s will become vectors and will be required for `calculate_weights`.

src/derivative_operators/BC_operators.jl
```diff
@@ -0,0 +1,149 @@
...
+
+        b_l = -αl[1]/denoml
+        b_r = -αr[1]/denomr
+        new{T, typeof(a_l)}(a_l,b_l,reverse!(a_r),b_r)
+    end
+end
+
+#implement Neumann and Dirichlet as special cases of RobinBC
+NeumannBC(α::AbstractVector{T}, dx::AbstractVector{T}, order) where T = RobinBC([zero(T), one(T), α[1]], [zero(T), one(T), α[2]], dx, order)
+NeumannBC(α::AbstractVector{T}, dx::AbstractVector{T}) where T = RobinBC([zero(T), one(T), α[1]], [zero(T), one(T), α[2]], dx)
```
ChrisRackauckas: There is no need for these dispatches that can be absorbed by default args.

ChrisRackauckas:
Not done, let's get that boundary sign sorted out.

xtalax:
@ajozefiak, how about changing the coeffs to α β γ as in ghost_node.pdf?

Here's my reasoning, at least for the a_l component - a_r is passing the tests as is. 
(1) αu[0] + βu'[0] = γ
With the first order forward difference approximation 
(2) (u[1]-u[0])/dx  ~ u'[0]

(2) -> (1)  => (3)
(3) αu[0] + β(u[1]-u[0])/dx = γ 
=  u[0](α - β/dx) + (β/dx)u[1] = γ
u[0] = (γ - (β/dx)u[1])/(α - β/dx)
so b_l = -(β/dx)/(α - β/dx)
multiply top and bottom by dx/β
b_l = -1/((αdx/β) -1)
or b_l = 1/(1- (αdx/β))

and in general 
(2') dot(s,u)/dx ~ u'[0]
(3')  αu[0] + (β/dx)dot(s,u) = γ 
= u[0](α + (β/dx)s[0]) + (β/dx)dot(s[1:end], u[1:end]) = γ 
u[0] = (γ - (β/dx)dot(s[1:end], u[1:end]))/(α + (β/dx)s[0])
so b_l = -(β/dx)s[1:end]/(α + (β/dx)s[0])
= -s[1:end]/(αdx/β + s[0])

And is that what is in the code? No it isn't... 
Really sorry about this guys, quite embarrassing - but I think this clears up the issue. PR incoming.

PR #108, opened 2019-06-15
fixed the sign errors

xtalax:
Tests all pass now, with the original test setup. (Sorry I ever doubted you @ajozefiak).

Take care that I've changed the convention back to [α, β, γ], so the new constructors will need to be merged as well.


ChrisRackauckas:
Doesn't look like tests pass. There's a sign error, but it's the right value.

ChrisRackauckas:
Cherrypicked.

PR #110, opened 2019-06-16
[WIP] Analytic tests for Derivative Operator and Fixing Stencils

ajozefiak:
I've implemented two analytically computed derivative operators for testing, derivative order = 2 and approximation order = 4, and derivative and approximation orders = 4. Currently the concretization tests against the analytic solutions are failing due to a mistake in DerivativeOperator. I am looking into this at the moment.


ajozefiak:
I've pushed the fixes to the boundary stencils in DerivativeOperator. Essentially we were calculating the boundary stencil weights with an offset that was one position too far to the left. Secondly, we weren't reversing the weights within each row for the high_boundary_coefs, instead we were only reversing the order of the rows (of the boundary coefs).

PR #117, opened 2019-06-17
Fix indexing

shivin9:
Avoiding matrix generation for indexing operations `A[i,:]` and `A[:,j]`.


src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -48,15 +48,32 @@ end
 @inline getindex(A::AbstractDerivativeOperator, ::Colon, ::Colon) = Array(A)
 
 @inline function getindex(A::AbstractDerivativeOperator, ::Colon, j)
-    return BandedMatrix(A)[:,j]
+    T = eltype(A.stencil_coefs)
```
ChrisRackauckas: Is this one actually more efficient? I'd like to see a quick benchmark.

shivin9: Yes it is faster.
``` julia
julia> function getindex1(A::AbstractDerivativeOperator, ::Colon, j)
           return BandedMatrix(A)[:,j]
           end
getindex1 (generic function with 1 method)

julia> @benchmark getindex1(A, :, 10)
BenchmarkTools.Trial: 
  memory estimate:  463.69 KiB
  allocs estimate:  10
  --------------
  minimum time:     148.569 μs (0.00% GC)
  median time:      168.989 μs (0.00% GC)
  mean time:        214.548 μs (18.84% GC)
  maximum time:     79.850 ms (99.72% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

vs the new approach

```julia
julia> @benchmark A[:,10]
BenchmarkTools.Trial: 
  memory estimate:  15.88 KiB
  allocs estimate:  2
  --------------
  minimum time:     12.657 μs (0.00% GC)
  median time:      16.986 μs (0.00% GC)
  mean time:        16.981 μs (0.00% GC)
  maximum time:     65.610 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

shivin9: @ChrisRackauckas I will try to remove BandedMatrices in other indexing operations also.

ChrisRackauckas: sounds like a good plan. That's a pretty dramatic speedup.

shivin9: I tried many different approaches but it seems that directly creating the full matrix and selecting a view is cheaper than copying vectors one by one.

PR #118, opened 2019-06-17
3rd order robin and general tests - Multidimensional BC first attempt

xtalax:
- Adds RobinBC and GeneralBC tests to 3rd approximation order
- Restructures BC types to allow for future extension, such as multiple layered boundaries.
- First pass of multidimensional BCs, 2D and 3D supported (Still needs concretization).
- Includes a general BoundaryPaddedArray to support the above, with some array like functions overloaded (still needs concretization and getindex Edit: now implemented).
- To generalize, the slicemul function could be moved to a tensor like operation, with repeat() of the BC stencil along the dimensions orthogonal to extension,  pointwise multiplication with the correct view of u, and contraction along the extended dimension at each boundary. Would likely require specialized MultiDimensional methods for each atomic BC type, or a restructuring of BCs to make clear where in u the stencil should be applied.


src/derivative_operators/BC_operators.jl
```diff
@@ -1,11 +1,15 @@
-abstract type AbstractBC{T} <: AbstractDiffEqLinearOperator{T} end
+abstract type AbstractBC{T} end#<: AbstractDiffEqLinearOperator{T} end
```
ChrisRackauckas: ?

xtalax: Ah that change is an error

src/derivative_operators/BC_operators.jl
```diff
@@ -1,11 +1,15 @@
...
+abstract type MultiDimensionalBC{T, N} <: AbstractBC{T} end
+abstract type AbstractBoundaryPaddedArray{T, N} <: AbstractArray{T, N} end
 """
 Robin, General, and in general Neumann and Dirichlet BCs are all affine opeartors, meaning that they take the form Qx = Qax + Qb.
 """
-abstract type AffineBC{T,V} <: AbstractBC{T} end
+abstract type AffineBC{T,V} <: SingleLayerBC{T} end
 
-struct PeriodicBC{T} <: AbstractBC{T}
+struct PeriodicBC{T} <: SingleLayerBC{T}
```
ChrisRackauckas: It can be `PeriodicBC{T,N}` for multi-dimensional. Then just work out how to build the padded arrays.

src/derivative_operators/BC_operators.jl
```diff
@@ -139,14 +145,150 @@ function LinearAlgebra.Array(Q::BoundaryPaddedVector)
...
+end
+
+
+"""
+Higher dimensional generalization of BoundaryPaddedVector, pads an array of dimension N with 2*N arrays of dimension N-1, stored in lower and upper.
+
+"""
+struct BoundaryPaddedArray{T, N, V <: AbstractArray{T}, B <: AbstractArray{T}} <: AbstractBoundaryPaddedArray{T, N}
+    lower::Vector{B}
+    upper::Vector{B}
```
ChrisRackauckas: for this, don't you need left right top bottom?

src/derivative_operators/BC_operators.jl
```diff
@@ -139,14 +145,150 @@ function LinearAlgebra.Array(Q::BoundaryPaddedVector)
...
+function Base.convert(::Type{AbstractMatrix},A::SingleLayerBC{T}) where T
     SparseMatrixCSC(A)
 end
+
+#######################################################################
+# Multidimensional
+#######################################################################
+
+# One of my utility functions, it works by applying an operator along a slice of u in the dimension specified by n
+# A multidim convolution would be better. - this is unused in DifEqOperators.jl but used as a template for the following functions
```
ChrisRackauckas: indeed, multidimensional convolutions (using NNLib.jl) is where we are going next on the GhostDerivativeOperators for `L1 + L2` combinations. But it shouldn't be necessary on the boundary.

src/derivative_operators/BC_operators.jl
```diff
@@ -139,14 +145,150 @@ function LinearAlgebra.Array(Q::BoundaryPaddedVector)
...
+            end
+        end
+    else
+        throw("Dim greater than 3 not supported!")
+    end
+    return lower, upper
+end
+
+struct MultiDimensionalSingleLayerBC{T, N} <: MultiDimensionalBC{T, N}
+    BC::Vector{SingleLayerBC{T}} # I think this has to be an array of non concrete BCs to allow different BCs on different dims
```
ChrisRackauckas: each point along the boundary can have a different BC, so maybe this needs to be an Vector{Array{SingleLayerBC{T},N-1}}. For example, 2D, [1] is a vector of BCs for the left/right BCs, and [2] is the vector of BCs for the top bottom. 3D, [1] is a matrix of left right BCs at each point in [y,z], [2] is matrix of top bottom bcs in [x,z], and [3] is a matrix of front back BCs in [x,y]. The higher dimensional extension of this interpretation is then clear.

src/derivative_operators/BC_operators.jl
ChrisRackauckas: Actually, could this get a similar N-dimensional interpretation as MultiDimensionalSingleLayerBC?

xtalax: This is N dimensional, lower is a vector of BCs to be applied to the lower index end, upper to the higher index end. Doesn't allow for mixed BCs along a boundary yet though

src/derivative_operators/BC_operators.jl
xtalax: That makes sense, I'll try to move towards this more flexible implementation next time i sit down with the code

src/derivative_operators/BC_operators.jl
xtalax: That would work well as a special case

xtalax: added

src/derivative_operators/BC_operators.jl
```diff
@@ -1,32 +1,35 @@
 abstract type AbstractBC{T} <: AbstractDiffEqLinearOperator{T} end
```
ajozefiak: We may want to get rid of this, in general our BC operators are affine as opposed to linear.

ajozefiak:
For simplicity, I am thinking that for higher dimensions it might make sense to try something at first like:

`struct RobinBC{T,N} <: AffineBC{T}
    a_l::Vector{T}
    b_l::T
    a_r::Vector{T}
    b_r::T
end`

and 

`struct BoundaryPaddedArray{T,N<:Int,M<:Int,V<:AbstractArray{T,M}} <: AbstractArray{T,M}
    l::Array{T,1}
    r::Array{T,1}
    u::V
end`



where `N` is the axis of differentiation and `M` is the dimension of the input array `U` and of the output `Q*U`. This sort of implementation follows the way we have set up the derivative operators and it should allow us to immediately consolidate the boundary operators with `GhostDerivativeOperator` in order to get a first pass at a functional higher-dimensional `LQ` operator.

src/derivative_operators/BC_operators.jl
xtalax: I don't remember adding it personally, happy to remove it. I think that AbstractBC should hook in to some unified type along with AbstractDerivativeOperator etc. somewhere to allow for later extension

xtalax:
> struct BoundaryPaddedArray{T,N<:Int,M<:Int,V<:AbstractArray{T,M}} <: AbstractArray{T,M}
>     l::Array{T,1}
>     r::Array{T,1}
>     u::V
> end

We could move to this sort of implementation - I would however like to move away from left/right as a notation because it becomes inconsistent and confusing at higher dimensions. Another thing is that the padding arrays should be in general of dimensionality M-1. There is a boundary padded array that already works, but it does contain all extensions at once.

> struct RobinBC{T,N} <: AffineBC{T}
>     a_l::Vector{T}
>     b_l::T
>     a_r::Vector{T}
>     b_r::T
> end

This becomes a little more problematic - Since we want to allow for mixed boundary conditions along a boundary, half dirichlet0 half neumann0 as a toy example, It becomes needlessly complex to give this functionality to the 'atomic' BCs, which only act in general on one vector like strip of a larger array. 
This is why I've opted to build some wrapper types - the MixedBC which allows you to use a different typed BCs on the lower and higher index end (This becomes important when you want to bridge two domains that are governed by different PDEs with a BridgeBC) and the MutiDimensionalSingleLayerBC, shown here:

```
struct MultiDimensionalSingleLayerBC{T<:Number, N, M} <: MultiDimensionalBC{T, N}
    BCs::Vector{UnionSingleLayerBCArray{T,M}} #The Vector has length N - one dimension M=N-1 array of BCs for each dimension
end
```
UnionSingleLayerBCArray{T,M} is a union of Array{SingleLayerBC{T},M} for all subtypes of SingleLayerBC{T}, which is needed to support different BC types along a boundary.

Each array holds the BCs that are to be applied along each 'strip' (think colon indexed) of `U` along a dimension, and there is one of these BC arrays for each dimension, held in the vector `BCs`. This already works for arbitrary BCs. Have a look in MultiDimBC_test.jl for an idea of how this can be used.

The kind of functionality you are talking about, where the user just wants the same RobinBC everywhere or along a single boundary and so shouldn't have to deal with building the BC arrays, can be implemented on top of this as a method of RobinBC which takes 'dim' as a keyword arg.

> where `N` is the axis of differentiation and `M` is the dimension of the input array `U` and of the output `Q*U`. This sort of implementation follows the way we have set up the derivative operators and it should allow us to immediately consolidate the boundary operators with `GhostDerivativeOperator` in order to get a first pass at a functional higher-dimensional `LQ` operator.

I think what you are saying here is that it would be easier to build the GhostDerivativeOperator if each part of a multi dimensional `Q` were split in to parts that acted along each dimension, like `Q = Qx*Qy*Qz`.  I think that is best achieved by giving MultiDimensionalSingleLayerBC an axis parameter, and splitting up the BCs Vector, further splitting the BoundaryPaddedArray as you suggest so that we don't have this vector for dims situation anywhere anymore - which eases the concretization of `LQ ` for ldiv. Would that be preferable?

Although looking at the state of ghost_derivative_operator.jl it seems that higher dimensional compositions might almost already work, needing just a generalized `mul!`, and `ldiv!` for `A`.  For `mul!`, one could simply write `mul!(temp_u, L, Array(Q*U))` and this would already work. `Q` would then just need a specialized ldiv! method, as avoiding concretization of the operator is important for large problem spaces for memory usage reasons. Then `A` wouldn't need to know about the implementation of `Q`, and could simply call its specialized methods. @ChrisRackauckas @ajozefiak what do you think about that?

ajozefiak:
I'll address some of the specific points in my morning, but yes, mul! should already work as we've implemented an arbitrary-dimension multiplication in `derivative_operator_function.jl`. However, this current implementation assumes that `Q*M` (where `M` is some array of arbitrary dimension) gives us a concrete type of `AbstractArray`. 

>I think what you are saying here is that it would be easier to build the GhostDerivativeOperator if each part of a multi dimensional Q were split in to parts that acted along each dimension, like Q = Qx*Qy*Qz. I think that is best achieved by giving MultiDimensionalSingleLayerBC an axis parameter, and splitting up the BCs Vector, further splitting the BoundaryPaddedArray as you suggest so that we don't have this vector for dims situation anywhere anymore - which eases the concretization of LQ for ldiv. Would that be preferable?
>

I'm not sure at the moment what would be more optimal, but a necessary implementation (and a reasonable first pass) would be to implement boundary operators that append an extra row (or slice if we're talking about higher dimensions) to each side of a particular axis. That is implementing `Qx`, `Qy`, `Qz`... so that if we have some 2-dimensional array M and we want to compute the second derivative along the second dimension, call it y, subject to boundary conditions encoded in `Qy` then we can call something like `Lyy * (Qy * M)`. 

The real question that arises is do we want to support (and in turn optimize around) `(Lxx+Lyy+Lzz)*Q` or `Lxx*Qx + Lyy*Qy + Lzz*Qz` or perhaps both will work equally well. 

Also, I was using the RobinBC as an example, we should generalize, but I just want to clear up implementation/interface decisions before diving too deep into generalizations.

xtalax:
@ajozefiak Goodnight :)

>  However, this current implementation assumes that `Q*M` (where `M` is some array of arbitrary dimension) gives us a concrete type of `AbstractArray`.

BoundaryPaddedArray <: AbstractBoundaryPaddedArray <:AbstractArray so that dispatch should still work.

> then we can call something like `Lyy * (Qy * M)`.

I think I'm seeing your point: No need to extend a dimension that we are not operating on.

> The real question that arises is do we want to support (and in turn optimize around) `(Lxx+Lyy+Lzz)*Q` or `Lxx*Qx + Lyy*Qy + Lzz*Qz` or perhaps both will work equally well.

I think that this depends on the implementation of  `L`, namely where we want to deal with any special boundary handling. At some point an improved `PeriodicBC` that takes multiple points over the boundary so that `L` is able to use it's standard stencil for all `U` will be preferable to implement, along with other BCs that can extend to multiple ghost nodes. Perhaps a specialized method of `convolve_bc_left/right!` could deal with this, dispatching on a `MultiLayerBoundaryPaddedArray`. 
With the split method it may also be possible to combine the BC stencil with the boundary stencil of `L` in the ghost operator, improving perfomance when the solve is iterated.

ChrisRackauckas:
>The real question that arises is do we want to support (and in turn optimize around) (Lxx+Lyy+Lzz)*Q or Lxx*Qx + Lyy*Qy + Lzz*Qz or perhaps both will work equally well.

We want to support both, but `(Lxx+Lyy+Lzz)*Q` is decisively better for cache efficiency with the convolutional kernels, and so that will be the preferred one in any case where speed is concerned, since the former can be done in n^3 calculations with efficient blocked kernels, while the latter takes 3n^3.

ChrisRackauckas:
>I think that this depends on the implementation of L, namely where we want to deal with any special boundary handling. At some point an improved PeriodicBC that takes multiple points over the boundary so that L is able to use it's standard stencil for all U will be preferable to implement, along with other BCs that can extend to multiple ghost nodes. Perhaps a specialized method of convolve_bc_left/right! could deal with this, dispatching on a MultiLayerBoundaryPaddedArray.

Yes indeed. The fast convolution kernels should have the boundary handling separate from the interior, and dispatch on whether it's a MultiLayerBoundaryPaddedArray or not. Then the interior is just the fast kernels, while the exterior might be a little messy.

xtalax:
> > The real question that arises is do we want to support (and in turn optimize around) (Lxx+Lyy+Lzz)_Q or Lxx_Qx + Lyy_Qy + Lzz_Qz or perhaps both will work equally well.
> 
> We want to support both, but `(Lxx+Lyy+Lzz)*Q` is decisively better for cache efficiency with the convolutional kernels, and so that will be the preferred one in any case where speed is concerned, since the former can be done in n^3 calculations with efficient blocked kernels, while the latter takes 3n^3.

Then I'll focus on the split Q implementation, with some added methods and types to allow convenient and fast extension of composed Q also

xtalax:
@ChrisRackauckas @ajozefiak  All functionality from this branch has been merged or superseded by that on SplitBC-dev, therefore I'm closing this pull request.

PR #121, opened 2019-06-19
Fixed convolutions for BoundaryPaddedVector + tests created

ajozefiak:


src/derivative_operators/convolutions.jl
```diff
@@ -65,45 +65,71 @@ function convolve_interior!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
...
     # Just do the middle parts
-    for i in (1+A.boundary_point_count) : (length(x_temp)-A.boundary_point_count)
+
+    # need to account for _x.l
+    i = 1 + A.boundary_point_count
+    xtempi = zero(T)
+    cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i-A.boundary_point_count] : stencil
+    cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : true
+    cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+    xtempi = cur_coeff*cur_stencil[1]*_x.l
```
ChrisRackauckas: This should be in `convolve_BC_left!`, and the other half in `convolve_BC_right!`. The reason is because we want to keep the interior, which is the bulk of the calculation, as simple as possible. This will allow us to specialize on it in the near future.

ajozefiak: Got it, I will make the fix

PR #122, opened 2019-06-20
Non equal spacing grid

shivin9:


ChrisRackauckas:
Needs a test which stresses the non-uniform part.

PR #123, opened 2019-06-23
[WIP] GhostDerivativeOperator \ implementation

ajozefiak:
Requires more tests for \

### DIFF
--- a/src/derivative_operators/ghost_derivative_operator.jl
+++ b/src/derivative_operators/ghost_derivative_operator.jl
@@ -23,8 +23,8 @@ function *(A::GhostDerivativeOperator{T,E,F}, M::AbstractMatrix{T}) where {T,E,F
 end
 
 function \(A::GhostDerivativeOperator{T,E,F}, u::AbstractVector{T}) where {T,E,F}
-    (QL,Qb) = sparse(A.Q, A.L.len)
-    return (A.L*QL) \ (u - A.L*Qb)
+    (AL,Ab) = Array(A)
+    return AL \ (u - Ab)
 end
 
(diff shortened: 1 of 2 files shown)


src/derivative_operators/ghost_derivative_operator.jl
```diff
@@ -12,18 +12,45 @@ function *(A::GhostDerivativeOperator{T,E,F}, u::AbstractVector{T}) where {T,E,F
     return A.L*(A.Q*u)
 end
 
+
+function *(A::GhostDerivativeOperator{T,E,F}, M::AbstractMatrix{T}) where {T,E,F}
```
ChrisRackauckas: still needs a non-allocating `mul!` overload

src/derivative_operators/ghost_derivative_operator.jl
```diff
@@ -12,18 +12,45 @@ function *(A::GhostDerivativeOperator{T,E,F}, u::AbstractVector{T}) where {T,E,F
...
+    @assert size(M,1) == A.L.len
+    M_temp = zeros(T, A.L.len, size(M,2))
+    for i in 1:size(M,2)
+        M_temp[:,i] = A.L*(A.Q*M[:,i])
+    end
+    return M_temp
+end
+
+function \(A::GhostDerivativeOperator{T,E,F}, u::AbstractVector{T}) where {T,E,F}
```
ChrisRackauckas: should make this fall back to an `ldiv!` overload.

ChrisRackauckas:
I think this is enough to merge. More tests and non-allocating versions should follow up.

PR #124, opened 2019-06-23
[WIP]: Upwind operators

shivin9:


src/derivative_operators/convolutions.jl
```diff
@@ -72,6 +83,7 @@ function convolve_interior!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
         @inbounds for idx in 1:A.stencil_length
             xtempi += cur_coeff * cur_stencil[idx] * x[(i-1) - (mid-idx) + 1]
+            @show i, idx, cur_stencil[idx], x[i-mid+idx]
```
ChrisRackauckas: oops

src/derivative_operators/convolutions.jl
```diff
@@ -83,45 +95,22 @@ function convolve_BC_left!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
...
         cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i] : true
+        # need to account for x.l in first interior
         xtempi = cur_coeff*cur_stencil[1]*_x.l
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
         @inbounds for idx in 2:A.boundary_stencil_length
             xtempi += cur_coeff * cur_stencil[idx] * _x.u[idx-1]
         end
         x_temp[i] = xtempi
     end
-    # need to account for x.l in first interior
```
ChrisRackauckas: why is this deleted?

src/derivative_operators/derivative_operator.jl
```diff
@@ -85,12 +103,13 @@ function UpwindDifference{N}(derivative_order::Int,
...
 
-    stencil_length          = derivative_order + approximation_order - 1 + (derivative_order+approximation_order)%2
+    stencil_length          = derivative_order + approximation_order
     boundary_stencil_length = derivative_order + approximation_order
-    dummy_x                 = -div(stencil_length,2) : div(stencil_length,2)
+    dummy_x                 = -1 : stencil_length - 2
     boundary_x              = -boundary_stencil_length+1:0
     boundary_point_count    = div(stencil_length,2) - 1 # -1 due to the ghost point
     # Because it's a N x (N+2) operator, the last stencil on the sides are the [b,0,x,x,x,x] stencils, not the [0,x,x,x,x,x] stencils, since we're never solving for the derivative at the boundary point.
+    @show dummy_x
```
ChrisRackauckas: this

ChrisRackauckas:
Needs tests

src/derivative_operators/convolutions.jl
shivin9: This code is not necessary. The upper part is doing the right thing.

src/derivative_operators/convolutions.jl
```diff
@@ -83,45 +95,22 @@ function convolve_BC_left!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
...
-    i = 1 + A.boundary_point_count
-    xtempi = zero(T)
-    cur_stencil = eltype(A.stencil_coefs) <: AbstractVector ? A.stencil_coefs[i-A.boundary_point_count] : A.stencil_coefs
-    cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : true
-    cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
-    xtempi = cur_coeff*cur_stencil[1]*_x.l
-    @inbounds for idx in 2:A.stencil_length
-        xtempi += cur_coeff * cur_stencil[idx] * x[(i-1) - (mid-idx) + 1]
-    end
-    x_temp[i] = xtempi
```
shivin9: This code is resetting `x_temp[i]` the wrong way. We should use `A.boundary_stencil_length` and not `A.stencil_length`.

src/derivative_operators/derivative_operator.jl
shivin9: When the stencil is not centred, it's length is different from the centred stencil. And since we want only 1 point on one side, I am starting `dummy_x` from `-1`.

shivin9:
Can you suggest some tests? I am calculating First derivatives etc. and that is working fine. Will push it soon.

ChrisRackauckas:
Second and third order upwinding is worked out here: https://en.wikipedia.org/wiki/Upwind_scheme#Third-order_upwind_scheme . A test that uses * for concretization and then tests the actual concretization against that would do well. Then write out what the third order one would be by hand, and make sure either way matches that.

src/derivative_operators/convolutions.jl
```diff
@@ -3,6 +3,11 @@ function LinearAlgebra.mul!(x_temp::AbstractVector{T}, A::DerivativeOperator, x:
     convolve_BC_left!(x_temp, x, A)
     convolve_interior!(x_temp, x, A)
     convolve_BC_right!(x_temp, x, A)
+    if typeof(A.dx) <: T
+        rmul!(x_temp, @.(1/(A.dx^A.derivative_order)))
+    else
+        nothing # what to do when dx is not uniform???
```
ChrisRackauckas: It should be baked into the stencil already.

shivin9: You mean that the onus is on the user to care of this part?

shivin9: How will be happen @ChrisRackauckas ? Any reference?

shivin9: is it element wise division only?

ajozefiak: If dx is not uniform then we are already accounting for the scaling when we compute the coefficients.

For instance, in the uniform grid case of [0.1, 0.2, 0.3, 0.4, ...] we compute the coefficients with respect to dx = 1 and then scale afterwards. 

In the case of a non-uniform grid such as [0.1, 0.19, 0.22, 0.25, ...] we are computing the coefficients with respect to a subset of the grid as it is, so the Fornberg algorithm will account for the 'scaling'.

ChrisRackauckas: We should probably do the same for uniform as well, but that can wait.

PR #126, opened 2019-06-28
Arbitrary axis of differentiation/convolutions

ajozefiak:
A first pass of a fall-back implementation of an arbitrary axis of differentiation. Still requires tests for higher dimensions, but existing tests for dimension N = 1 are passing.

### DIFF
--- a/src/derivative_operators/derivative_operator_functions.jl
+++ b/src/derivative_operators/derivative_operator_functions.jl
@@ -1,40 +1,32 @@
-for MT in [AbstractArray, AbstractMatrix]
-    @eval begin
-        function LinearAlgebra.mul!(x_temp::$MT{T}, A::DerivativeOperator{T,N}, M::$MT{T}) where {T<:Real,N}
+function LinearAlgebra.mul!(x_temp::AbstractArray{T}, A::DerivativeOperator{T,N}, M::AbstractArray{T}) where {T<:Real,N}
 
-            v = zeros(ndims(x_temp))
-            v[N] = 2
-            @assert [size(x_temp)...]+v == [size(M)...]
+    v = zeros(ndims(x_temp))
+    v[N] = 2
+    @assert [size(x_temp)...]+v == [size(M)...]
 
-            ndimsM = ndims(M)
-            @assert N <= ndimsM
+    ndimsM = ndims(M)
+    @assert N <= ndimsM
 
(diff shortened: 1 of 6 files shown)


src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -0,0 +1,40 @@
+for MT in [AbstractArray, AbstractMatrix]
```
ChrisRackauckas: Why have a separate one for AbstractMatrix? AbstractMatrix == AbstractArray{T,2}

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -0,0 +1,40 @@
...
+            setindex!(idx, :, N)
+            for I in indices
+                Base.replace_tuples!(nidx, idx, idx, otherdims, I)
+                mul!(view(x_temp, idx...), A, view(M, idx...))
+            end
+        end
+    end
+end
+
+for MT in [AbstractVector, AbstractMatrix, AbstractArray]
```
ChrisRackauckas: just AbstractArray here

src/derivative_operators/derivative_operator_functions.jl
ajozefiak: I was getting ambiguity issues with 
`mul!(x_temp::AbstractArray{T,2}, A::AbstractDerivativeOperator{T}, M::AbstractMatrix{T})`

ajozefiak: Since `AbstractDerivativeOperator{T}` doesn't specify the axis of differentiation I had to specify `A` as being a `DerivativeOperator{T,N}`

ajozefiak: I could use something like `typeof(A).parameters[2]` for `mul!(x_temp::AbstractArray{T}, A::AbstractDerivativeOperator{T}, M::AsbtractArray{T}) where T`, but I don't know if this is okay with the generality of `AbstractDerivativeOperator` that we might want to maintain.

ajozefiak:
As suggested, redundancy of types has been addressed. To resolve ambiguities I've changed `AbstractVecOrMat` to `AbstractArray` in `common_defaults.jl` along with removing subsumed `mul!` and `*` methods in `abstract_operator_functions.jl`.

PR #127, opened 2019-06-30
Add `opnorm` to jacvec operators

YingboMa:
This PR enables `JacVecOperator` to work with `Exprb` solvers.

### DIFF
--- a/src/matrixfree_operators.jl
+++ b/src/matrixfree_operators.jl
@@ -6,7 +6,7 @@ mutable struct MatrixFreeOperator{F,N,S,O} <: AbstractMatrixFreeOperator{F}
   opnorm::O
   ishermitian::Bool
   function MatrixFreeOperator(f::F, args::N;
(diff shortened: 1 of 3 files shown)


src/matrixfree_operators.jl
```diff
@@ -6,7 +6,7 @@ mutable struct MatrixFreeOperator{F,N,S,O} <: AbstractMatrixFreeOperator{F}
   opnorm::O
   ishermitian::Bool
   function MatrixFreeOperator(f::F, args::N;
-                              size=nothing, opnorm=nothing, ishermitian=false) where {F,N}
+                              size=nothing, opnorm=true, ishermitian=false) where {F,N}
```
ChrisRackauckas: why is this true/false?

YingboMa: `opnorm` is defaulted to be one, as `opnorm` is only used to test relative tolerance in `expv`.

PR #128, opened 2019-07-02
mul! implemented for 2d and 3d multiplication with NNlib

ajozefiak:
The implementation could be made cleaner, but the general working idea is present in my branch.


ChrisRackauckas:
Is there a test which hits this dispatch?

ajozefiak:
I'll need to add a fewmore tests, but the tests I've created for the fallback in `differentiation_dimension.jl` hit the 3D case along with the fallback for higher dimensions. I will need to explicitly test the 2D case, but that is also being hit (for differentiation along rows) in a few places in `derivative_operator_interface.jl`

PR #129, opened 2019-07-03
Fixed dispatch 2D/3D multiplication for DerivativeOperator

ajozefiak:
As title, along with adding a few tests to make sure to cover a few previously uncovered cases.

### DIFF
--- a/src/derivative_operators/derivative_operator_functions.jl
+++ b/src/derivative_operators/derivative_operator_functions.jl
@@ -27,8 +27,8 @@ end
 
 for MT in [2,3]
     @eval begin
-        function LinearAlgebra.mul!(x_temp::AbstractArray{T,$MT}, A::DerivativeOperator{T,N,Wind,T2,S1}, M::AbstractArray{T,$MT}) where {T<:Real,N,Wind,T2,SL,S1<:SArray{Tuple{SL},T,1,SL}}
-
+        function LinearAlgebra.mul!(x_temp::AbstractArray{T,$MT}, A::DerivativeOperator{T,N,false,T2,S1}, M::AbstractArray{T,$MT}) where {T<:Real,N,Wind,T2,SL,S1<:SArray{Tuple{SL},T,1,SL}}
+            
             # Check that x_temp has correct dimensions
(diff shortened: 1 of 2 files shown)


src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -38,52 +38,49 @@ for MT in [2,3]
             ndimsM = ndims(M)
```
ChrisRackauckas: set this to dispatch only when dowind is false

ajozefiak: @ChrisRackauckas dispatching on false now, should be good to merge once tests pass

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -38,52 +38,49 @@ for MT in [2,3]
...
-            for I in indices
-                Base.replace_tuples!(nidx, idx, idx, otherdims, I)
-                convolve_BC_left!(view(x_temp, idx...), view(M, idx...), A)
-                convolve_BC_right!(view(x_temp, idx...), view(M, idx...), A)
+            if bpc > 0
+                dimsM = [axes(M)...]
+                alldims = [1:ndims(M);]
+                otherdims = setdiff(alldims, N)
+
+                idx = Any[first(ind) for ind in axes(M)]
```
ChrisRackauckas: Why `Any`?

ChrisRackauckas:
Looks good, one last comment.

src/derivative_operators/derivative_operator_functions.jl
ajozefiak: @ChrisRackauckas at line 76 I am calling setindex!(idx, :, N), and Colon is not a subtype of Number

ChrisRackauckas: oh yes.

ChrisRackauckas: This might ruin inference though. Maybe `ntuple` is more appropriate? That can be optimized later

PR #130, opened 2019-07-03
Upwind Operators

shivin9:


src/derivative_operators/convolutions.jl
```diff
@@ -3,7 +3,6 @@ function LinearAlgebra.mul!(x_temp::AbstractVector{T}, A::DerivativeOperator, x:
     convolve_BC_left!(x_temp, x, A)
     convolve_interior!(x_temp, x, A)
     convolve_BC_right!(x_temp, x, A)
-    rmul!(x_temp, @.(1/(A.dx^A.derivative_order)))
```
ChrisRackauckas: Make a quick separate PR that does this so it doesn't mix with the winding one. That should be quick and easy.

shivin9:
Anything missing?

ChrisRackauckas:
Is this still WIP?

src/derivative_operators/convolutions.jl
```diff
@@ -12,14 +12,23 @@ function convolve_interior!(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::
...
     for i in (1+A.boundary_point_count) : (length(x_temp)-A.boundary_point_count)
         xtempi = zero(T)
         cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i-A.boundary_point_count] : stencil
         cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i] : coeff isa Number ? coeff : true
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
         for idx in 1:A.stencil_length
-            xtempi += cur_coeff * cur_stencil[idx] * x[i - mid + idx]
+            x_idx = use_winding(A) && cur_coeff < 0 ? x[i + mid - idx] : x[i - mid + idx]
+            xtempi += cur_coeff * cur_stencil[idx] * x_idx
+            # @show i, idx, cur_stencil[idx], i-mid+idx, x[i-mid+idx]
```
ChrisRackauckas: delete

src/derivative_operators/convolutions.jl
```diff
@@ -28,12 +37,27 @@ end
 function convolve_BC_left!(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::DerivativeOperator) where {T<:Real}
     stencil = A.low_boundary_coefs
     coeff   = A.coefficients
-    for i in 1 : A.boundary_point_count
-        cur_stencil = stencil[i]
+
+    _bpc = A.boundary_point_count
+    flag = false
+    if stencil == []
```
ChrisRackauckas: isempty

ChrisRackauckas: if it's empty, wouldn't you just return?

src/derivative_operators/convolutions.jl
```diff
@@ -43,34 +67,65 @@ end
...
 end
 
 ###########################################
 
 # Against A BC-padded vector, specialize the computation to explicitly use the left, right, and middle parts
 function convolve_interior!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector, A::DerivativeOperator) where {T<:Real}
+    @assert length(x_temp) == length(_x.u)
     stencil = A.stencil_coefs
     coeff   = A.coefficients
-    x = _x.u
```
ChrisRackauckas: ?

src/derivative_operators/convolutions.jl
```diff
@@ -43,34 +67,65 @@ end
...
-    for i in (2+A.boundary_point_count) : (length(x_temp)-A.boundary_point_count)-1
+    for i in (1+A.boundary_point_count) : (length(x_temp)-A.boundary_point_count)
         xtempi = zero(T)
         cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i-A.boundary_point_count] : stencil
         cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : coeff isa Number ? coeff : true
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
         @inbounds for idx in 1:A.stencil_length
-            xtempi += cur_coeff * cur_stencil[idx] * x[(i-1) - (mid-idx) + 1]
+            # @show i, idx, cur_stencil[idx], i-mid+idx, x[i-mid+idx]
+            x_idx = use_winding(A) && cur_coeff < 0 ? _x[i + mid - idx] : _x[i - mid + idx]
```
ChrisRackauckas: don't directly index _x.

test/bc_coeff_compositions.jl
```diff
@@ -216,21 +220,21 @@ end
     ghost_u = A \ u
 
     # Check that A\u.(x) is consistent with analytic_AL \ u.(x)
-    @test analytic_u ≈ ghost_u
+    @test_broken analytic_u ≈ ghost_u
```
ChrisRackauckas: adding upwinding should definitely not break these parts.

shivin9:
> Is this still WIP?

Its complete from my side. Just needs some more tests.

test/bc_coeff_compositions.jl
shivin9: This should be fixed by #145.

src/derivative_operators/convolutions.jl
shivin9: Taken cared of in #145.

src/derivative_operators/convolutions.jl
shivin9: No, we have to still take care of the boundary point which and instead of boundary stencil, use the normal interior stencil.

src/derivative_operators/convolutions.jl
shivin9: ?

ChrisRackauckas: this comment. I assume you meant to put this as no longer WIP?

test/upwind_operators_interface.jl
```diff
@@ -0,0 +1,226 @@
+using SparseArrays, DiffEqOperators, LinearAlgebra, Random,
```
ChrisRackauckas: this file needs to get added to runtests.jl

test/derivative_operators_interface.jl
```diff
@@ -125,7 +131,7 @@ end
     @test Array(A) ≈ correct
     @test sparse(A) ≈ correct
     @test BandedMatrix(A) ≈ correct
-    @test A*y ≈ Array(A)*y
+    @test A*y ≈ Array(A)*y atol=0.1
```
ChrisRackauckas: why was this added and what does this have to do with upwinding?

src/derivative_operators/concretization.jl
```diff
@@ -4,18 +4,26 @@ function LinearAlgebra.Array(A::DerivativeOperator{T}, N::Int=A.len) where T
...
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(A.low_boundary_coefs[i]) : A.low_boundary_coefs[i]
         L[i,1:bstl] = cur_coeff * cur_stencil
     end
+
     for i in bl+1:N-bl
         cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i] : coeff isa Number ? coeff : true
         stencil     = eltype(A.stencil_coefs) <: AbstractVector ? A.stencil_coefs[i] : A.stencil_coefs
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(stencil) : stencil
-        L[i,i+1-stl_2:i+1+stl_2] = cur_coeff * cur_stencil
+        L[i,i+1-stl_2:i-stl_2+stl] = cur_coeff * cur_stencil
```
ChrisRackauckas: why would this change in the non-winding case as well?

src/derivative_operators/convolutions.jl
```diff
@@ -43,92 +66,165 @@ end
...
 function convolve_interior!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector, A::DerivativeOperator) where {T<:Real}
+    @assert length(x_temp) == length(_x.u)
     stencil = A.stencil_coefs
     coeff   = A.coefficients
+
+    mid = div(A.stencil_length,2)
+
     x = _x.u
-    mid = div(A.stencil_length,2) + 1
+    i = 1+A.boundary_point_count
```
ChrisRackauckas: this seems like boundary handling?

src/derivative_operators/convolutions.jl
```diff
@@ -43,92 +66,165 @@ end
...
         cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
         @inbounds for idx in 1:A.stencil_length
-            xtempi += cur_coeff * cur_stencil[idx] * x[(i-1) - (mid-idx) + 1]
+            x_idx = use_winding(A) && cur_coeff < 0 ? x[i + mid - idx - 1] : x[i - mid + idx - 1]
+            xtempi += cur_coeff * cur_stencil[idx] * x_idx
         end
         x_temp[i] = xtempi
     end
+
+    # The right boundary part
```
ChrisRackauckas: this also seems like boundary handling?

test/bc_coeff_compositions.jl
```diff
@@ -77,7 +81,7 @@ end
         mul!(view(LQM,:,i), L, Q*M[:,i])
     end
     ghost_LQM = A*M
-    @test ghost_LQM ≈ LQM
+    @test ghost_LQM ≈ LQM atol=0.25 # unacceptable level of tolerance
```
ChrisRackauckas: why did the tolerance change here?

test/derivative_operators_interface.jl
shivin9: Fixed after #145.

src/derivative_operators/concretization.jl
ChrisRackauckas: Bump

ChrisRackauckas: Bump

src/derivative_operators/convolutions.jl
```diff
@@ -43,92 +66,189 @@ end
...
+            cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : coeff isa Number ? coeff : true
+            cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+            @inbounds for idx in 1:A.stencil_length
+                x_idx = use_winding(A) && cur_coeff < 0 ? x[i + mid - idx - 1] : x[i - mid + idx - 1]
+                xtempi += cur_coeff * cur_stencil[idx] * x_idx
+            end
+            x_temp[i] = xtempi
+        end
+
+        # The right boundary part
```
ChrisRackauckas: this shouldn't be in here...

src/derivative_operators/convolutions.jl
```diff
@@ -43,92 +66,189 @@ end
...
+        i = 1 + A.boundary_point_count
+        idx = 1
+        xtempi = zero(T)
+        cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i-A.boundary_point_count] : stencil
+        cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : coeff isa Number ? coeff : true
+        cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+        # @show cur_stencil
+        x_idx = _x.l
+        xtempi += cur_coeff * cur_stencil[idx] * x_idx
+        @inbounds for idx in 2:A.stencil_length
```
ChrisRackauckas: why is this here?

shivin9:
I have separated the handling of Upwind and normal convolutional operations
as writing a generalized code was becoming too terse.

shivin9:
It may seem that this can be outside the loop but this part of code varies
for the different cases. Therefore we need separate declarations.

ChrisRackauckas:
`DerivativeOperator{T,N,true}` vs `DerivativeOperator{T,N,false}` dispatches

src/derivative_operators/convolutions.jl
```diff
@@ -43,92 +66,195 @@ end
...
+    i = 1 + A.boundary_point_count
+    idx = 1
+    xtempi = zero(T)
+    cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i-A.boundary_point_count] : stencil
+    cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i-A.boundary_point_count] : coeff isa Number ? coeff : true
+    cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+    # @show cur_stencil
+    x_idx = _x.l
+    xtempi += cur_coeff * cur_stencil[idx] * x_idx
+    @inbounds for idx in 2:A.stencil_length
```
ChrisRackauckas: this stuff should not be added to this convolution.

shivin9:
Isn't it necessary? This is for explicitly handling the ghost point, the
_x.l. Otherwise, we will have to do _x[1] which is inefficient.

ChrisRackauckas:
you should never do the ghost point in the interior... otherwise you violate the assumptions of constant stencils and https://github.com/JuliaDiffEq/DiffEqOperators.jl/pull/132 will never work.

shivin9:
Stencils are always constant in convolve_interior!, its just that there arises a case when the interior stencil touches the ghost point at the end. The point for which that happens, I am taking in as an interior point instead of a ghost point.

`[g,x1,x2,x3,x4,x5,x6,x7,g]`
if interior stencil length is 5, then basically x2 is handled by convolve_interior with the interior stencil .

ChrisRackauckas:
>Stencils are always constant in convolve_interior!, its just that there arises a case when the interior stencil touches the ghost point at the end.

Which is why that's not in the interior...

ChrisRackauckas:
If x2 is in the interior, then there would be no point to be handled on the left since `x.l` is the only point to the left of that. That makes no sense.

shivin9:
That depends on the definition of interior-point right? According to what I understand if a point cannot be processed using an interior stencil then it is a boundary point.

x2 is essentially using the interior stencil only(in the implementation) so its an interior point.

Still, if you want other way then I have written it also, it's just breaking many tests.

ChrisRackauckas:
All of the points which are kept are interior points. The convolutions don't need to even happen at the points on the boundary. So interior means that points which don't touch the boundary, since that's required for the convolution to be constant in order to use cudnn...

src/derivative_operators/concretization.jl
ChrisRackauckas: Bump

src/derivative_operators/concretization.jl
```diff
@@ -31,17 +39,26 @@ function SparseArrays.SparseMatrixCSC(A::DerivativeOperator{T}, N::Int=A.len) wh
...
```
ChrisRackauckas: this too

src/derivative_operators/convolutions.jl
```diff
@@ -28,12 +36,27 @@ end
 function convolve_BC_left!(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::DerivativeOperator) where {T<:Real}
     stencil = A.low_boundary_coefs
     coeff   = A.coefficients
-    for i in 1 : A.boundary_point_count
-        cur_stencil = stencil[i]
+
+    _bpc = A.boundary_point_count
+    flag = false
```
ChrisRackauckas: what exactly does this flag do? Name it.

src/derivative_operators/derivative_operator.jl
```diff
@@ -24,16 +24,25 @@ function CenteredDifference{N}(derivative_order::Int,
     stencil_length          = derivative_order + approximation_order - 1 + (derivative_order+approximation_order)%2
     boundary_stencil_length = derivative_order + approximation_order
     dummy_x                 = -div(stencil_length,2) : div(stencil_length,2)
-    boundary_x              = -boundary_stencil_length+1:0
     boundary_point_count    = div(stencil_length,2) - 1 # -1 due to the ghost point
+
+    left_boundary_x         = 0:(boundary_stencil_length-1)
```
ChrisRackauckas: why is the centered difference operator changed in this PR?

shivin9: This was a part of the fix of #145 when I merged that branch. Without this other tests will break.

ChrisRackauckas: Why would other tests break if they weren't broken before?

shivin9: Well they were not testing the required things... like the new error_benchmark test which Alex wrote or the mistake in the stencil which was not summing upto 0. I don't know how but the `ldiv!` tests were passing before.

src/derivative_operators/convolutions.jl
shivin9: Ok. This is when there are no boundary points.

src/derivative_operators/derivative_operator.jl
ChrisRackauckas: What does this have to do with upwinding?

shivin9: It was a part of#145 which got merged. I merged it as it had changes relevant to upwind operators also.

ChrisRackauckas: can we have upwinding go in without all of this, and separate it out so I can see the diff? There still hasn't been an upwinding PR...

shivin9: Ok.

shivin9:
Closed in favour of #155.

PR #131, opened 2019-07-04
Stencils takes care of scaling factor

shivin9:


ChrisRackauckas:
Looks good but tests are failing. The concretizations probably have to be changed as well.

shivin9:
What has to be changed in the concretization?

ChrisRackauckas:
They probably had to scale as well in order to be correct, but now they don't.

PR #132, opened 2019-07-06
2d multiplication for combinations

ajozefiak:
This branch is complete in terms of functionality, however the implementation could be refactored. Additionally, this branch only implements the 2D case, the 3D case can also be hardcoded (easily) following the final implementation of this branch. 

To summarize: This branch implements mul! for compositions of DerivativeOperators and makes a single efficient call to NNlib.conv!, provided there are some operators in the composition that:
1) operate on a regular grid, i.e `dx isa Number`
2) do not use upwinding
3) have a scalar coefficient (or no coefficient) 

For all other operators in the composition that do not satisfy the above conditions, the fallback mul! method is invoked.

I believe that I have complete test coverage for the 2D case, though there may be some cases that I have overlooked. Additionally, my tests test that multiplication with compositions agrees with performing multiplication on each individual operator and then summing the outputs. That is, I am testing for cases like: `(Lxx+Lyy)*M ≈ Lxx*M + Lyy*M`. Provided that multiplication is implemented correctly, our convolutions are implemented correctly, and that our stencils for the derivative operators are computed correctly, then the successful passing of this branch's tests is an indication of the correctness of its implementation. 

Additionally, this implementation assumes that if we have a composition of operators `A` that contains at least one operator operating on the first axis and at least one operator operating on the second axis. Then `A*M` will map a `(N+2)x(M+2)` matrix to a `NxM` matrix. Otherwise, only the axis on which all operators are operating on will be truncated in length.


test/bc_coeff_compositions.jl
```diff
@@ -197,19 +197,19 @@ end
 @testset "Test Left Division L4 (fourth order)" begin
 
     # Test \ homogenous and inhomogenous BC
-    dx = 0.00001
-    x = 0.0001:dx:0.01
+    dx = 0.01
+    x = 0.0001:dx:0.2
```
ChrisRackauckas: why is the left end so many zeros if dx = 0.01?

ajozefiak: I'll fix that, I forgot to change the left end.

src/derivative_operators/convolutions.jl
```diff
@@ -132,3 +132,70 @@ function convolve_BC_right!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
         x_temp[bc_start + i] = xtempi
     end
 end
+
+###########################################
+
+# Implementations of additive convolutions, necessary for compositions of operators
```
ChrisRackauckas: what is an additive convolution? This just looks like a convolution?

src/derivative_operators/convolutions.jl
```diff
@@ -132,3 +132,70 @@ function convolve_BC_right!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
...
+        cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i] : coeff isa Number ? coeff : true
+        cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+        for idx in 1:A.stencil_length
+            xtempi += cur_coeff * cur_stencil[idx] * x[i - mid + idx]
+        end
+        x_temp[i] += xtempi
+    end
+end
+
+function convolve_interior_add_range!(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::DerivativeOperator, offset::Int) where {T<:Real}
```
ChrisRackauckas: This could be done with views?

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -106,3 +141,239 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
...
 end
+
+
+###########################################
+
+# A more efficient mul! implementation for compositions of operators which may include regular-grid, centered difference,
+# scalar coefficient, non-winding, DerivativeOperator, operating on a 2D or 3D AbstractArray
+function LinearAlgebra.mul!(x_temp::AbstractArray{T,2}, A::AbstractDiffEqCompositeOperator, M::AbstractArray{T,2}) where {T}
+
+    # opsA operators satisfy conditions for NNlib.conv! call, opsB operators do not
```
ChrisRackauckas: then use a more informative name

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -106,3 +141,239 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
...
+    opsA = DerivativeOperator[]
+    opsB = DerivativeOperator[]
+    for L in A.ops
+        if (L.coefficients isa Number || L.coefficients === nothing) && use_winding(L) == false && L.dx isa Number
```
ChrisRackauckas: this should be generalized to a trait on the operators

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -106,3 +141,239 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
...
+        idx = div.(Wdims,2).+1
+
+        # add to kernel each stencil
+        for L in opsA
+            s = L.stencil_coefs
+            sl = L.stencil_length
+            axis = typeof(L).parameters[2]
+            offset = convert(Int64,(Wdims[axis] - sl)/2)
+            coeff = L.coefficients isa Number ? L.coefficients : true
+            for i in offset+1:Wdims[axis]-offset
```
ChrisRackauckas: I'm not seeing how this accounts for the dimension of the operator?

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -25,6 +26,38 @@ function LinearAlgebra.mul!(x_temp::AbstractArray{T}, A::DerivativeOperator{T,N}
...
+
+    # Check that axis of differentiation is in the dimensions of M and x_temp
+    ndimsM = ndims(M)
+    @assert N <= ndimsM
+
+    dimsM = [axes(M)...]
+    alldims = [1:ndims(M);]
+    otherdims = setdiff(alldims, N)
+
+    idx = Any[first(ind) for ind in axes(M)]
```
ChrisRackauckas: there must be a better way to type this?

src/derivative_operators/convolutions.jl
```diff
@@ -132,3 +132,70 @@ function convolve_BC_right!(x_temp::AbstractVector{T}, _x::BoundaryPaddedVector,
...
+    mid = div(A.stencil_length,2)
+    for i in [(1+A.boundary_point_count):(A.boundary_point_count+offset); (length(x_temp)-A.boundary_point_count-offset+1):(length(x_temp)-A.boundary_point_count)]
+        xtempi = zero(T)
+        cur_stencil = eltype(stencil) <: AbstractVector ? stencil[i] : stencil
+        cur_coeff   = typeof(coeff)   <: AbstractVector ? coeff[i] : coeff isa Number ? coeff : true
+        cur_stencil = use_winding(A) && cur_coeff < 0 ? reverse(cur_stencil) : cur_stencil
+        for idx in 1:A.stencil_length
+            xtempi += cur_coeff * cur_stencil[idx] * x[i - mid + idx]
+        end
+        x_temp[i] += xtempi
```
ChrisRackauckas: oh it's because of the `+=`? That seems like it could be a keyword argument in the standard kernels then, rather than a copy-paste

ChrisRackauckas:
In general, this is looking very good. However, there are some code re-use issues I think should get addressed.

src/derivative_operators/derivative_operator_functions.jl
ajozefiak: Here I'm using "dimension of the kernel" to mean the length of the interior stencils and ultimately the minimum area required for the convolution kernel to contain the sum of all of the interior stencils.

src/derivative_operators/convolutions.jl
```diff
@@ -21,11 +21,11 @@ function convolve_interior!(x_temp::AbstractVector{T}, x::AbstractVector{T}, A::
         for idx in 1:A.stencil_length
             xtempi += cur_coeff * cur_stencil[idx] * x[i - mid + idx]
         end
-        x_temp[i] = xtempi
+        x_temp[i] = xtempi + !:($overwrite)*x_temp[i]
```
ChrisRackauckas: that's a weird way to write `!overwrite * x_temp[i]`

ChrisRackauckas:
```suggestion
        x_temp[i] = xtempi + !overwrite*x_temp[i]
```

ajozefiak: Yes... I should change that. I think I was originally trying to do something more convoluted with metaprogramming but forgot to get rid of the inference.

ChrisRackauckas:
Merging to get this going along, but two comments left should be addressed in a future PR.

PR #134, opened 2019-07-07
Fix kernel flipping in NNlib.conv call

ajozefiak:
@ChrisRackauckas  this should be good to merge. The only change is that we are now flipping the kernel correctly before calling NNlib.conv!. Also, a few tests for non-symmetric stencils were added, which were missing earlier and hence we didn't see this issue earlier.


ChrisRackauckas:
Good catch.

PR #137, opened 2019-07-10
Stencil scaling for regular grids.

ajozefiak:
Stencil scaling for regular grids is now performed in the constructor for CenteredDifference as opposed to scaling the stencils every time we concretize, multiply, or perform left division. A single test for left division was modified to use a smaller dx due to the problem being quite unstable, causing machine errors to propagate.

### DIFF
--- a/test/bc_coeff_compositions.jl
+++ b/test/bc_coeff_compositions.jl
@@ -198,12 +198,12 @@ end
 
     # Test \ homogenous and inhomogenous BC
     dx = 0.01
-    x = 0.0001:dx:0.2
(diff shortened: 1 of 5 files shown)


test/bc_coeff_compositions.jl
```diff
@@ -197,19 +197,19 @@ end
 @testset "Test Left Division L4 (fourth order)" begin
 
     # Test \ homogenous and inhomogenous BC
-    dx = 0.00001
-    x = 0.0001:dx:0.01
+    dx = 0.01
```
ChrisRackauckas: same here

PR #138, opened 2019-07-13
Higher Dimension Concretization

ajozefiak:
Implementation of higher (arbitrary) dimension concretization. 

The concretization is implemented such that given an operator `L`, operating on an arbitrary axis, and an AbstractArray `M` on which `L` is operating on. Then `Array(L, size(M))` will produce a `BandedBlockBandedMatrix` which operators on a vector of length `length(vec(M))`, i.e flattened `M`. 

Likewise, `sparse` and `BandedMatrix` are also implemented, each of which producing a `BandedBlockBandedMatrix`, though differing in the concretization of `L` in the Kron product used to construct `sparse(L, size(M))` and `BandedMatrix(L, size(M))`.


src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -91,3 +91,44 @@ function *(A::DerivativeOperator{T,N},M::AbstractArray{T}) where {T<:Real,N}
     LinearAlgebra.mul!(x_temp, A, M)
     return x_temp
 end
+
+function LinearAlgebra.ldiv!(M_temp::AbstractArray{T,MT}, A::DerivativeOperator{T,N}, M::AbstractArray{T,MT}) where {T<:Real, N, MT}
```
xtalax: Could I suggest a change throughout this function? It will improve memory performance if you concretize the operators to SparseArrays, as memory can be quite a major limitation on your domain size. 

If A and B have N^2 elements each, Kron(A,B) will have N^4. But since A and B are in this case sparse Multidiagonal,  if you use this kind of concretization you can reduce the memory consumption of kron(A,B) to O(N^2) since they have O(N) elements in sparse form.

ajozefiak: Yes, that is what I will go towards.

ChrisRackauckas: I think as a first pass to get tests working it's fine to have it concrete, but I agree that in any case this will need to be sparse.

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -106,3 +106,80 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
...
+            end
+        end
+
+        # compute ldiv!
+        ldiv!(vec(M_temp), factorize(Array(B)), vec(M))
+    end
+end
+
+# Non-inplace left division.
+function \(A::DerivativeOperator{T,N}, M::AbstractArray{T,MT}) where {T<:Real, N,MT}
```
ChrisRackauckas: this can just allocate and call the first?

ajozefiak: I don't think there's no a `ldiv!` implementation for sparse matrices in general (according to what I could find in `SparseArrays.jl`). So from my understanding, if we want to exploit the sparse structure of our operators we should be using `\`.

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -106,3 +106,80 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
         A.low_boundary_coefs,
         A.high_boundary_coefs,coefficients,A.coeff_func)
 end
+
+# Inplace left division
+function LinearAlgebra.ldiv!(M_temp::AbstractArray{T,MT}, A::DerivativeOperator{T,N}, M::AbstractArray{T,MT}) where {T<:Real, N, MT}
```
ChrisRackauckas: Shouldn't this just be a concretization function, nd ldiv should just call the concretization function? I don't think we need `\` overloads, just good concretizers on the composed methods?

ajozefiak: @ChrisRackauckas So just scrap `ldiv!` and `\` and instead implement things like `Array` and `SparseMatrixCSC` for higher dimensional operators?

ChrisRackauckas: Yeah that would work.

src/derivative_operators/concretization.jl
```diff
@@ -95,3 +95,202 @@ end
...
+    inds = Array(1:N)
+    inds[1], inds[D] = inds[D], inds[1]
+    Q_b = [permutedims(add_dims(bc_tuple[2], N-1), inds) for bc_tuple in bc_tuples]
+
+    return (Q_L, Q_b)
+end
+
+SparseArrays.sparse(Q::MultiDimDirectionalBC, N) = SparseMatrixCSC(Q, N)
+
+function BandedMatrices.BandedMatrix(Q::MultiDimDirectionalBC{T, B, D, N, K}, M) where {T, B, D,N,K}
```
ChrisRackauckas: these will usually be a BlockBandedMatrix?

src/derivative_operators/concretization.jl
```diff
@@ -247,3 +247,91 @@ LinearAlgebra.Array(Q::ComposedMultiDimBC, Ns) = Tuple(Array.(Q.BCs, Ns))
...
+
+    # Case where A is differentiating along hte first dimension
+    else
+        n = 1
+        for M_i in Mshape[2:end]
+            n *= M_i
+        end
+        B = Kron(Eye(n), Array(A))
+    end
+    return BandedBlockBandedMatrix(B)
```
ChrisRackauckas: This should return an Array

src/derivative_operators/concretization.jl
```diff
@@ -247,3 +247,91 @@ LinearAlgebra.Array(Q::ComposedMultiDimBC, Ns) = Tuple(Array.(Q.BCs, Ns))
...
+
+    # Case where A is differentiating along hte first dimension
+    else
+        n = 1
+        for M_i in Mshape[2:end]
+            n *= M_i
+        end
+        B = Kron(sparse(I,n,n), sparse(A))
+    end
+    return BandedBlockBandedMatrix(B)
```
ChrisRackauckas: Should return a SparseMatrixCSC

ajozefiak: Can it still be a lazy array? As in a Kron product of SparseMatrixCSC's or must it completely be a SparseMatrixCSC?

ChrisRackauckas: We want to return the array type the user asks for if they use the constructor. This is because they make assumptions that only hold for specific array types (which is why they directly called the constructor). The form that we should be smart with is the `sparse` function.

PR #140, opened 2019-07-17
Scalar coefficient handling added. 

ajozefiak:
This PR adds scalar coefficient handling. This helps with the mul! dispatch for the 2D/3D case.

### DIFF
--- a/src/derivative_operators/derivative_operator_functions.jl
+++ b/src/derivative_operators/derivative_operator_functions.jl
@@ -57,7 +57,7 @@ for MT in [2,3]
             W = zeros(Wdims...)
             Widx = Any[Wdims...]
             setindex!(Widx,:,N)
-            coeff = A.coefficients === Nothing ? True : A.coefficients
+            coeff = A.coefficients === nothing ? True : A.coefficients
             W[Widx...] = s
 
             cv = DenseConvDims(_M, W, padding=pad,flipkernel=true)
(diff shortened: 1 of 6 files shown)


src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -27,8 +27,8 @@ end
 
 for MT in [2,3]
     @eval begin
-        function LinearAlgebra.mul!(x_temp::AbstractArray{T,$MT}, A::DerivativeOperator{T,N,false,T2,S1}, M::AbstractArray{T,$MT}) where {T<:Real,N,T2,SL,S1<:SArray{Tuple{SL},T,1,SL}}
-
+        function LinearAlgebra.mul!(x_temp::AbstractArray{T,$MT}, A::DerivativeOperator{T,N,false,T2,S1,S2,T3}, M::AbstractArray{T,$MT}) where
```
ChrisRackauckas: I don't think we use S2 or T3?

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -27,8 +27,8 @@ end
...
+                                                                            {T<:Real,N,T2,SL,S1<:SArray{Tuple{SL},T,1,SL},S2,T3<:Union{Nothing,Number}}
```
ChrisRackauckas: Why <: Real? That would break complex.

Issue #148, opened 2019-08-06
Boundary value operators: final touchups


ChrisRackauckas:
For the most part the boundary value operators are looking good after https://github.com/JuliaDiffEq/DiffEqOperators.jl/pull/135 . Some finishing details are:

- [x] Let Robin accept abstract vector
- [x] Simplified constructors don't need to allocate a vector
- [x] slicemul docstring space
- [ ] dim type parameter?

xtalax:
Dim is a type parameter of MultiDimDirectionalBC - do you mean that it should be declared as such with a `where {dim}` in those constructors?

xtalax:
@ChrisRackauckas solved in #149

ChrisRackauckas:
Still the type parameter issue

xtalax:
I was getting confusing errors trying to get it to work with a type parameter like that - since `MultiDimBC` is not a type, just a function name.

ChrisRackauckas:
You need to fake it being a type.

jagot:
> [x] Let Robin accept abstract vector

I wondered about this point; the constructor for the uniform case only accepts `Tuple`s whereas the non-uniform takes `AbstractVector`s. `GeneralBC` takes `AbstractVector`s for both cases.

ChrisRackauckas:
Completed by https://github.com/SciML/DiffEqOperators.jl/pull/279

PR #149, opened 2019-08-06
#148 Changes implemented

xtalax:
- Let Robin accept abstract vector
- Let General accept abstract vector
- Simplified constructors to take tuples rather than vectors
- Updated slice_rmul docstring
- Split out some utility functions in to their own file

Did not change dim in to type parameter - `MultiDimBC` is not a type, just a unified constructor name for all types of multi dimensional boundary condition - so julia gets confused when you try to give it a type parameter. Even an extra constructor for `MultiDimDirectionalBC{dim}` was causing problems when called by a  method of `MultiDimBC`, with julia thinking that `dim` was supposed to be used as `T` for some reason. 

Not sure if i'm being obtuse, or if this should be posted to #gripes :stuck_out_tongue_closed_eyes:


src/derivative_operators/BC_operators.jl
```diff
@@ -37,24 +37,24 @@ end
...
-    function RobinBC(l::AbstractVector{T}, r::AbstractVector{T}, dx::T, order = 1) where {T}
+    function RobinBC(l::NTuple{3,T}, r::NTuple{3,T}, dx::T, order = 1) where {T}
         αl, βl, γl = l
         αr, βr, γr = r
 
         s = calculate_weights(1, one(T), Array(one(T):convert(T,order+1))) #generate derivative coefficients about the boundary of required approximation order
 
-        a_l = -s[2:end]./(αl*dx/βl + s[1])
-        a_r = s[end:-1:2]./(αr*dx/βr - s[1]) # for other boundary stencil is flippedlr with *opposite sign*
+        a_l = SVector{length(s)-1}(-s[2:end]./(αl*dx/βl + s[1]))
```
ChrisRackauckas: this won't infer

src/utils.jl
```diff
@@ -0,0 +1,80 @@
...
+        if !b
+            return false
+        end
+    end
+    return true
+end
+
+"""
+A function to generate the correct permutation to flip an array of dimension `N` to be orthogonal to `dim`, when the Nth dimension is a singleton
+"""
```
ChrisRackauckas: I don't get this

xtalax: it's useful if you have an Array{T,N-1} `A` that you want to concatanate with an Array{T,N} `B`.
You need to give `A` an extra dimension of length 1 so that it's `N` dimensional, and then permute its dimensions such that it is pointed orthogonally to the dimension that you want to join it to `B` along, i.e. the length 1 dimension is in position `dim`.
```
A = rand(2,6,7)
B = rand(2,5,6,7) #we want to join A to be along dimension 2
A_N = add_dims(A, 1)

A_permuted = permutedims(A_N, orth_perm(4, 2))
cat(A_permuted, B, dims=2)
```

src/derivative_operators/BC_operators.jl
xtalax: Can you think of a way to get it to infer?

ChrisRackauckas: it just won't like these. This constructor should make vectors. If we made the Fornberg function pure then it could.

src/utils.jl
ChrisRackauckas: so it's just reshape?

xtalax: Well, when you put it that way :sweat_smile:...
I hadn't realised that this was equivalent

src/utils.jl
```diff
@@ -0,0 +1,65 @@
...
+Efficiently reduce a BitArray with or (|)
+"""
+function or(B::BitArray)
```
ChrisRackauckas: `all(B)` and `!all(B)`. Use the standard iterators from Base instead of renaming them

xtalax: It seems I need to sit down and read the base documentation end to end - these do exactly the same thing?

ChrisRackauckas: yes.

Issue #151, opened 2019-08-07
2D multiplication touchups


ChrisRackauckas:
https://github.com/JuliaDiffEq/DiffEqOperators.jl/pull/132#discussion_r310369525

https://github.com/JuliaDiffEq/DiffEqOperators.jl/pull/132#discussion_r310369533

still need to be addressed, but weren't blocking.

ChrisRackauckas:
Also we need to make sure that GhostDerivative operators are handled correctly in this.

PR #152, opened 2019-08-07
[WIP] 3D mul! for compositions

ajozefiak:
The implementation here will follow from the 2D case. I will touch up both implementations according to #151 afterwards.


src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -347,3 +347,282 @@ function LinearAlgebra.mul!(x_temp::AbstractArray{T,2}, A::AbstractDiffEqComposi
...
+                    end
+                end
+            end
+            # convolve boundaries and unaccounted for interior in axis 2
+            if length(ops_2) > 0
+                for i in 1:size(x_temp)[1]
+                    for j in 1:size(x_temp)[3]
+                        # in the case of no axis 1 operators, we need to overwrite x_temp
+                        if length(ops_1) == 0
+                            convolve_BC_left!(view(x_temp,i,:,j), view(M,i+offset_x,:,j+offset_z), opsA[ops_2_max_bpc_idx...])
```
ChrisRackauckas: this is getting gnarly, but it looks right?

ajozefiak:
It's still WIP because I've just copied the 2D mul! and managed to thus far modify some parts of it. There are a few parts that I still need to modify for the 3D case.

ChrisRackauckas:
Let's merge and continue finding out what's up with that one case. It'll be good to have these tests tracked for everything else.

PR #156, opened 2019-08-23
Write a README

ChrisRackauckas:


README.md
```diff
@@ -5,7 +5,211 @@
...
+α[1] + α[2]u[0] + α[3]u'[0] + α[4]u''[0]+... = 0
+
+```julia
+GeneralBC(αl::AbstractArray{T}, αr::AbstractArray{T}, dx::AbstractArray{T}, order = 1)
+```
+
+### Multidimensional Boundary Conditions
+
+```julia
+Q_dim = MultiDimBC(Q, size(u), dim)
```
ChrisRackauckas: To be consistent with the derivative operators, this should be `MultiDimBC{dim}(Q, size(u))` @xtalax

README.md
```diff
@@ -5,7 +5,211 @@
...
+to apply the same boundary values to all dimensions, one can use
+
+```julia
+Qx,Qy,Qz = MultiDimBC(YourBC, size(u)) # Here u is 3d
+```
+
+Multidimensional BCs can then be composed into a single operator with:
+
+```julia
+Q = compose(BCs...)
```
ChrisRackauckas: @xtalax I think we could just overload `+` to compose? `Q = Qx + Qy + Qz` seems nicer.

PR #170, opened 2019-09-03
Fixes #168 - Multidimensional handling for GhostDerivativeOperator

### LINKED ISSUE #168: Missing Method + for GhostDerivativeOperator
From what I can tell so far there is currently no way to impose (Periodic, I haven't tested the other BCs) BC on problems which require adding multiple DerivativeOperators. One possibility to solve this might be to overload + for GhostDerivativeOperators?

xtalax:
Fixes #168, but reveals more work to do before `GhostDerivativeOperator` is complete.
#TODO
- [x] Full flat matrix concretization of multidimensional BCs
- [x] Concretizations of composite operators that take size
- [x] General `GhostDerivativeOperator` methods that work with `u::AbstractArray{T,N}`
- [x] Tests that check this


xtalax:
@ajozefiak 
This attempted fix has revealed some assumptions in `GhostDerivativeOperator` that we are still working in 1D. I'd appreciate it if you could take a look at this.

Especially wondering about the `*(A::GhostDerivativeOperator, u::AbstractMatrix)` method that's currently in `master`.
It seems to be assuming that `A` always acts along dimension 1, unless I'm missing something - Is this correct?

ajozefiak:
I'll have a look soon. The 1D dimensional assumption is wrong, that was implemented earlier in order to sync with the 1D BC's we had. We didn't have the chance to implement higher dimensional GhostDerivativeOperator functionality.

ajozefiak:
The other comment I will add is that I don't know if we should have an explicit `ldiv!` implementation. For the higher dimensional `DerivativeOperators` I left out an `ldiv!` implementation and instead implemented only various concretizations. This allows the user to specify which concretization to use in a solver. I believe for `GhostDerivativeOperator` we should follow a similar path.

xtalax:
I have added a fallback `ldiv!` and `\` that uses the full concretization, but will leave it at that for now. This is something that can be revisited at some point

src/derivative_operators/concretization.jl
```diff
@@ -383,4 +521,49 @@ function BlockBandedMatrices.BandedBlockBandedMatrix(A::DerivativeOperator{T,N},
...
+    return (Array(A.L,N)*Array(A.Q,A.L.len)[1], Array(A.L,N)*Array(A.Q,A.L.len)[2])
+end
+
+function LinearAlgebra.Array(A::GhostDerivativeOperator{T, E, F}, s::NTuple{N,I}) where {T,E,F,N,I<:Int}
+    return (Array(A.L, s)*Array(A.Q, s)[1], Array(A.L, s)*Array(A.Q, s)[2])
+end
+
+
+function BandedMatrices.BandedMatrix(A::GhostDerivativeOperator{T, E, F},N::Int=A.L.len) where {T,E,F}
+    return (BandedMatrix(A.L,N)*Array(A.Q,A.L.len)[1], BandedMatrix(A.L,N)*Array(A.Q,A.L.len)[2])
```
ChrisRackauckas: I don't think this will produce banded matrices

ChrisRackauckas:
I think it's fine to have the concretization just use `sparse` since that's our best guess at "the best one". We just need to make sure it's really the case.

src/derivative_operators/concretization.jl
```diff
@@ -383,4 +523,49 @@ function BlockBandedMatrices.BandedBlockBandedMatrix(A::DerivativeOperator{T,N},
...
+    return (SparseMatrixCSC(A.L,N)*SparseMatrixCSC(A.Q,A.L.len)[1], SparseMatrixCSC(A.L,N)*SparseMatrixCSC(A.Q,A.L.len)[2])
+end
+
+function SparseArrays.SparseMatrixCSC(A::GhostDerivativeOperator{T, E, F}, s::NTuple{N,I}) where {T,E,F,N,I<:Int}
+    return (SparseMatrixCSC(A.L,s)*SparseMatrixCSC(A.Q,s)[1], SparseMatrixCSC(A.L,s)*SparseMatrixCSC(A.Q,s)[2])
+end
+
+
+function SparseArrays.sparse(A::GhostDerivativeOperator{T, E, F}, N::Int=A.L.len) where {T,E,F}
+    return SparseMatrixCSC(A,N)
```
ChrisRackauckas: this should usually be a banded matrix, but not when periodic BCs.

test/3D_laplacian.jl
```diff
@@ -1,21 +1,36 @@
-using DiffEqOperators
-
+using DiffEqOperators, DifferentialEquations, ProgressMeter
+function main()
```
ChrisRackauckas: why is this in a function?

test/3D_laplacian.jl
```diff
@@ -1,21 +1,36 @@
-using DiffEqOperators
-
+using DiffEqOperators, DifferentialEquations, ProgressMeter
```
ChrisRackauckas: you don't need the full package here, at most OrdinaryDiffEq

test/3D_laplacian.jl
```diff
@@ -1,21 +1,36 @@
...
+t = 0.0:dt:10/3e8
+
+f(u,p,t) = (3e8)^2 .*(A*Q*u)
+
+uolder = deepcopy(u0)
+uold = deepcopy(u0)
+u = deepcopy(u0)
+
+function steptime(u,uold,uolder)
+    return ((dt^2 .*f(u,0,0) .+ 5u .- 4uold .+ uolder)./2, u, uold)
```
ChrisRackauckas: we need to add this method :(

xtalax: Do we still need this method? If so I'll try write something general like the RobinBC constructor. 

Isn't this just an Integrator though?


-

test/3D_laplacian.jl
xtalax: to avoid the problem of global variables being mutated in a loop not working

test/3D_laplacian.jl
ChrisRackauckas: The issue is that we're missing this IMEX method in OrdinaryDiffEq. It would be good to add.

ChrisRackauckas:
So where are we at here? Can you give a summary?

xtalax:
Im working with @grahamas to rebase this PR to the current master, much of the necessary code is written and working with just a few stubborn (and in one notable case downright bizarre) tests to get passing. There's still no block banded concretization but all other concretizations are written and passing.

ChrisRackauckas:
That sounds fantastic! Thanks for the update.

bosonbaas:
My group is very interested in using multidimensional BCs with `DiffEqOperators.jl`, and would be happy to contribute to this PR in any way we can! Are there any specific areas we could help out in?

grahamas:
@bosonbaas I'm working my way through resolving testing errors in #246. If you'd like to help out, you could work ahead of me on tests I haven't gotten to yet. If that sounds good to you, just pull down `grahamas:local-rebased` (which is ahead of the PR branch since I don't own the PR branch). I"m currently working through tests in `multi_dim_bc_test.jl`, so you could just comment out that line in `runtests.jl` and then run `]test`.  There's at least one failure down in `BasicSDOExamples.jl`.

That's the only thing I can think of! It's also the only thing I'm even aware of, so someone else who has a broader scope might have a better suggestion.

ChrisRackauckas:
I think that's a good place to work. I don't like really long PRs since it tends to bog things down, so this whole style of having a rebase branch and then trying to do everything in one PR is really what's prolonging the change. But since this is so far along, collaborating on the rebased branch is probably the quickest way to get this done.

ChrisRackauckas:
This was all included in https://github.com/SciML/DiffEqOperators.jl/pull/279 . Thanks to everyone who helped get this done!

Issue #182, opened 2019-09-21
Banded concretization of derivatives with boundary conditions gives dense matrices


jagot:
Consider the following example:

```julia
julia> L = 10.0
10.0

julia> N = 9
9

julia> δx = L/(N+1)
1.0

julia> Δ = CenteredDifference(2, 2, δx, N)
DerivativeOperator{Float64,1,false,Float64,StaticArrays.SArray{Tuple{3},Float64,1,3},StaticArrays.SArray{Tuple{0},StaticArrays.SArray{Tuple{4},Float64,1,4},1,0},Nothing,Nothing}(2, 2, 1.0, 9, 3, [1.0, -2.0, 1.0], 4, 0, StaticArrays.SArray{Tuple{4},Float64,1,4}[], StaticArrays.SArray{Tuple{4},Float64,1,4}[], nothing, nothing)

julia> BandedMatrix(Δ)
9×11 BandedMatrix{Float64,Array{Float64,2},Base.OneTo{Int64}}:
 1.0  -2.0   1.0   0.0   0.0    ⋅     ⋅     ⋅     ⋅     ⋅    ⋅
 0.0   1.0  -2.0   1.0   0.0   0.0    ⋅     ⋅     ⋅     ⋅    ⋅
 0.0   0.0   1.0  -2.0   1.0   0.0   0.0    ⋅     ⋅     ⋅    ⋅
 0.0   0.0   0.0   1.0  -2.0   1.0   0.0   0.0    ⋅     ⋅    ⋅
 0.0   0.0   0.0   0.0   1.0  -2.0   1.0   0.0   0.0    ⋅    ⋅
  ⋅    0.0   0.0   0.0   0.0   1.0  -2.0   1.0   0.0   0.0   ⋅
  ⋅     ⋅    0.0   0.0   0.0   0.0   1.0  -2.0   1.0   0.0  0.0
  ⋅     ⋅     ⋅    0.0   0.0   0.0   0.0   1.0  -2.0   1.0  0.0
  ⋅     ⋅     ⋅     ⋅    0.0   0.0   0.0   0.0   1.0  -2.0  1.0
```
(Should not the bandwidths be smaller?)

If I now add BCs, I get a dense matrix upon concretization:
```julia
julia> Q = Dirichlet0BC(typeof(δx))
RobinBC{Float64,Array{Float64,1}}([-0.0, 0.0], 0.0, [-0.0, 0.0], 0.0)

[...]

PR #183, opened 2019-09-22
Concretization cleanup

jagot:
Before starting to look at #182, I refactored the concretization, since so much was in common between the different array types.

Impacts #173.

### DIFF
--- a/src/derivative_operators/concretization.jl
+++ b/src/derivative_operators/concretization.jl
@@ -1,123 +1,60 @@
-function LinearAlgebra.Array(A::DerivativeOperator{T}, N::Int=A.len) where T
-    L = zeros(T, N, N+2)
+function Base.copyto!(L::AbstractMatrix{T}, A::DerivativeOperator{T}, N::Int) where T
     bl = A.boundary_point_count
     stencil_length = A.stencil_length
+    stencil_pivot = use_winding(A) ? (1 + stencil_length%2) : div(stencil_length,2)    
     bstl = A.boundary_stencil_length
+    
     coeff   = A.coefficients
-
-    if use_winding(A)
-        stencil_pivot = 1 + A.stencil_length%2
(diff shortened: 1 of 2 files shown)


src/derivative_operators/concretization.jl
```diff
@@ -1,123 +1,60 @@
...
+LinearAlgebra.Array(A::DerivativeOperator{T}, N::Int=A.len) where T =
+    copyto!(zeros(T, N, N+2), A, N)
 
-    if use_winding(A)
-        stencil_pivot = 1 + A.stencil_length%2
-    else
-        stencil_pivot = div(stencil_length,2)
-    end
+SparseArrays.SparseMatrixCSC(A::DerivativeOperator{T}, N::Int=A.len) where T =
+    copyto!(spzeros(T, N, N+2), A, N)
```
ChrisRackauckas: This would be really slow?

jagot: Why would it? I overloaded `Base.copyto!` at the top of the file, the code of which is exactly the same as was in all three versions for different target matrices.

ChrisRackauckas: Adding to a sparse matrix using indexing is really slow: the indices should all be declared up front.

ChrisRackauckas: oh I see, ehh, it's not any worse than before, but it's still bad. If no one wants to actually do the `(I,J,V)`, then it's best to probably just build the banded matrix and then convert.

jagot: Sure :) I was anyway interested in the banded versions, but could not keep my hands off refactoring this.

ChrisRackauckas: Yeah, 99% of the time people should just call `sparse` and get the banded version.

Issue #176, opened 2019-09-11
Interface for when the Jacobian is known analytically


rveltz:
If I have a vector field `f!(du,u,p,t)` and its differential at `u` applied to `v` given by `df!(du,u,v,p,t)`. How can I use `DiffEqOperators` to pass the analytical jacobian? I know I can do
```
ff = ODEFunction(f!, jac_prototype = JacVecOperator{Float64}(f!,x0,p))
```
but I am interested in the case where I have an expression for the Jacobian.

It seems it should be a modification of the code for `JacVecOperator`. Is there any other way?

Thank you for your help,

ChrisRackauckas:
You can use the MatrixFreeOperator: https://github.com/JuliaDiffEq/DiffEqOperators.jl/blob/master/src/matrixfree_operators.jl

rveltz:
Sorry but I don't understand what you suggest. To me `MatrixFreeOperator` does not provide a way to say that the jacobian is evaluated at `u` on tangent vector `v`. 

> A MatrixFreeOperator is a linear operator A*u where the action of A is explicitly defined by an in-place function f(du, u, p, t).

The definition of my linear operator is `v->df!(du,u,v,p,t)` has more arguments than above.

ChrisRackauckas:
In this setup, `update_coefficients!` would be used to update the values in `p` so that the operator becomes `J(u,p,t)`, and then it will perform `J(u,p,t)*v`. However, I can see how this is far less than ideal, so we should probably create an operator to simplify this a bit.

rveltz:
You mean that I need to create a struct to record `u,p`? It seems that `JacVecOperator` has more or less this interface

ChrisRackauckas:
Yeah, that's what `JacVecOperator` is doing. It just records the pointers and then uses it. So this just needs a similar wrapper, but without the AD part.

rveltz:
it is pretty simple actually but one has to modify `JacVecOperator` to add `df` and parametrise the type of `autodiff` so that df is called when `typeof(autodiff) != Bool` using dispach for example. If this is done I could plug `ODEProblem` in [PseudoArcLengthContinuation.jl](https://github.com/rveltz/PseudoArcLengthContinuation.jl). I would need to extract the linearsolver though which does not seem incoded in `ODEProblem`

PR #184, opened 2019-09-22
add analyticaljacvec operator

ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/DiffEqOperators.jl/issues/176 @rveltz try it out?

### DIFF
--- a/src/DiffEqOperators.jl
+++ b/src/DiffEqOperators.jl
@@ -46,7 +46,7 @@ for T in [DiffEqScaledOperator, DiffEqOperatorCombination, DiffEqOperatorComposi
 end
 
 export MatrixFreeOperator
-export JacVecOperator, getops
+export AnalyticalJacVecOperator, JacVecOperator, getops
 export AbstractDerivativeOperator, DerivativeOperator,
        CenteredDifference, UpwindDifference
 export DirichletBC, Dirichlet0BC, NeumannBC, Neumann0BC, RobinBC, GeneralBC, MultiDimBC, PeriodicBC,
--- a/src/jacvec_operators.jl
+++ b/src/jacvec_operators.jl
@@ -132,3 +132,52 @@ function LinearAlgebra.mul!(du::AbstractVector,L::JacVecOperator,x::AbstractVect
         end
     end
 end
+
+### AnalyticalOperator Implementation
+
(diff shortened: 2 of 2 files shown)


rveltz:
Hi,

It works but I had to correct a few typos. Should I paste the corrected code `jacvec_operators.jl` in here?

Thank you a lot for your work to making this happens.

> I want to use stiff solvers on GPU, I still have to implement a linear solver in your interface

ChrisRackauckas:
>I want to use stiff solvers on GPU, I still have to implement a linear solver in your interface

The default linear solver should use a QR that works?

rveltz:
I have to write a wrap to `KrylovKit` as you did for `LinearSolver`. What about my question about `jacvec_operators.jl`?

ChrisRackauckas:
Yes, please PR to this PR if you have the stuff to finish it!

>I have to write a wrap to KrylovKit as you did for LinearSolver.

`LinSolveGMRES` doesn't work with GPUs? If not, I think it would be good to add the KrylovKit stuff to DiffEqBase with an `@requires` and get it documented.

rveltz:
`LinSolveGMRES doesn't work with GPUs?` I think so. It is just that I use `KrylovKit` in my work for several reasons as I exposed on discourse.

As for the PR, I will try my best. A PR to a PR for my very first PR seems engaging :D

ChrisRackauckas:
>LinSolveGMRES doesn't work with GPUs? I think so. It is just that I use KrylovKit in my work for several reasons as I exposed on discourse.

Interesting, can you link to it? I haven't found benchmarks between the two. It would likely be very beneficial to add this `ComposePreconditioner` by default to KrylovKit as well:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/linear_nonlinear.jl#L146-L221

rveltz:
OK...I messed up :( I think you can see my PR [here](https://github.com/rveltz/DiffEqOperators.jl/pull/1)

The discourse post is [here](https://discourse.julialang.org/t/nonlinear-equations-on-abstract-vectors/28692/5)

I think `IterativeSolvers` is a bit faster than `KrylovKit`. We should probably add `struct LinKrylovKit` first.

ChrisRackauckas:
>We should probably add struct LinKrylovKit first.

Yes, let's continue this.

rveltz:
> Yes, let's continue this.

I messed up again but [see](https://github.com/rveltz/DiffEqBase.jl/pull/1)

PR #191, opened 2019-10-28
add and update documentation

thisrod:
This is what I've figured out so far.  Please correct anything that I'm mistaken about.


src/docstrings.jl
```diff
@@ -0,0 +1,154 @@
+@doc """
```
ChrisRackauckas: these should probably just be added directly to the function definitions, but this is fine.

ChrisRackauckas:
I'm not sure about what you're saying on the multidimensional BCs, but they probably do need to be worked out more. But we can continue that in another PR, and maybe that's the remaining issue.

PR #192, opened 2019-10-29
Add poisson equation example

jw3126:
If you like the example, I will try a few more. We could later add them via Literate.jl to the docs.


ChrisRackauckas:
Looks like a great example! But tests fail.

examples/heat_equation.jl
```diff
@@ -0,0 +1,36 @@
...
+const Δ = CenteredDifference(ord_deriv, ord_approx, h, nknots)
+const bc = Dirichlet0BC(Float64)
+
+t0 = 0.0
+t1 = 0.03
+u0 = u_analytic.(knots, t0)
+
+step(u,p,t) = Δ*bc*u
+prob = ODEProblem(step, u0, (t0, t1))
+alg = KenCarp4(autodiff=false)
```
ChrisRackauckas: why autodiff=false?

jw3126: If I turn it to true, I get:
```julia
MethodError: no method matching *(::GhostDerivativeOperator{Float64,DerivativeOpera
tor{Float64,1,false,Float64,StaticArrays.SArray{Tuple{3},Float64,1,3},StaticArrays.SArray{Tuple{0},St
aticArrays.SArray{Tuple{4},Float64,1,4},1,0},Nothing,Nothing},RobinBC{Float64,Array{Float64,1}}}, ::A
rray{ForwardDiff.Dual{ForwardDiff.Tag{DiffEqDiffTools.UDerivativeWrapper{ODEFunction{false,typeof(ste
p),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing
,Nothing},Float64,DiffEqBase.NullParameters},Float64},Float64,12},1})
```

jw3126:
I would like to add a simple transport equation example. For instance the steady state of transport in the unit interval with a constant velocity field. Roughly this should be:
```julia
using DiffEqOperators

nknots = 10
h = 1.0/(1+nknots)
knots = range(h, step=h, length=nknots)

D = UpwindDifference(1,1,h,nknots, _ -> 1.0)
bc = DirichletBC(1.0, 1.0)
(D *bc) \ zeros(nknots)
```
However I only want to specify the inflow, which is enough in theory. Is it possible to specify only one boundary with `DiffEqOperators`?

examples/heat_equation.jl
ChrisRackauckas: Try https://github.com/JuliaDiffEq/DiffEqOperators.jl/pull/193

ChrisRackauckas:
>However I only want to specify the inflow, which is enough in theory. Is it possible to specify only one boundary with DiffEqOperators?

No because the sizes don't line up. However, it won't actually use one of the boundary points in the multiplication, so the result is unchanged.

ChrisRackauckas:
Rebase and see if autodiff works

jw3126:
> > However I only want to specify the inflow, which is enough in theory. Is it possible to specify only one boundary with DiffEqOperators?
> 
> No because the sizes don't line up. However, it won't actually use one of the boundary points in the multiplication, so the result is unchanged.

Mhh I think it does use both boundaries:
```julia
using DiffEqOperators

nknots = 5
h = 1.0/(1+nknots)
knots = range(h, step=h, length=nknots)

D = UpwindDifference(1,1,h,nknots, _ -> 1.0)
bc = DirichletBC(10.0, 20.0)
(D *bc) \ zeros(nknots)
```
```
5-element Array{Float64,1}:
 10.0
 10.0
 10.0
 20.0
 20.0
```
And even if it worked, then that would be an implementation detail of `\` and could fail for other solvers.

ChrisRackauckas:
Let's continue that discussion

jw3126:
Why is the first order `UpwindDifference` size `(N, N+2)` instead of `(N, N+1)`? Is the only motivation to match the size of padded vectors? 
If so, why not make boundary conditions one-sided? For instance, advection-diffusion would look like this:
```julia
div = UpwindDifference(...) # size (N, N+1)
Δ = CenteredDiffernce(...)  # size (N, N+2)
Qleft = Dirichlet(10.0, side=:lower)  # size (N+1, N)
Qright = Dirichlet(20.0, side=:upper) # size (N+1, N)
Q = Qleft * Qright # size (N+2, N)

D = div*Qleft + Δ*Q
```

ChrisRackauckas:
That depends on the coefficient though. If your coefficient is time-varying and switching between positive and negative, then you'd be switching for whether you need a left bc or not. So it's more so what the choice of coefficient is that determines whether the left/right BC is needed, but it should always have a possible definition?

PR #213, opened 2019-12-09
Use `prod` intrinsic

charleskawczynski:
- Trim some lines using `prod` intrinsic

### DIFF
--- a/src/derivative_operators/concretization.jl
+++ b/src/derivative_operators/concretization.jl
@@ -220,25 +220,16 @@ BandedMatrices.BandedMatrix(Q::ComposedMultiDimBC, Ns) = Tuple(BandedMatrix.(Q.B
 function LinearAlgebra.Array(A::DerivativeOperator{T,N}, Mshape) where {T,N}
     # Case where A is not differentiating along the first dimension
     if N != 1
-        n = 1
-        for M_i in Mshape[1:N-1]
-            n *= M_i
-        end
+        n = prod(Mshape[1:N-1])
         B = Kron(Array(A), Eye(n))
         if N != length(Mshape)
-            n = 1
-            for M_i in Mshape[N+1:end]
-                n *= M_i
-            end
+            n = prod(Mshape[N+1:end])
             B = Kron(Eye(n), B)
         end
 
     # Case where A is differentiating along the first dimension
     else
-        n = 1
-        for M_i in Mshape[2:end]
-            n *= M_i
-        end
+        n = prod(Mshape[2:end])
         B = Kron(Eye(n), Array(A))
     end
     return Array(B)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Nice! It would be good to add `@view` in there to reduce the temporaries.

src/derivative_operators/concretization.jl
```diff
@@ -220,25 +220,16 @@ BandedMatrices.BandedMatrix(Q::ComposedMultiDimBC, Ns) = Tuple(BandedMatrix.(Q.B
 function LinearAlgebra.Array(A::DerivativeOperator{T,N}, Mshape) where {T,N}
     # Case where A is not differentiating along the first dimension
     if N != 1
-        n = 1
-        for M_i in Mshape[1:N-1]
-            n *= M_i
-        end
+        n = prod(Mshape[1:N-1])
```
ChrisRackauckas:
```suggestion
        n = prod(@view(Mshape[1:N-1]))
```

charleskawczynski:
Ah, `Mshape` is not always an `AbstractArray`
```
First Dimension: Error During Test at /home/travis/build/JuliaDiffEq/DiffEqOperators.jl/test/concretization.jl:18
  Test threw exception
  Expression: L1 * M ≈ Array(L1, size(M)) * vec(M) ≈ sparse(L1, size(M)) * M ≈ BandedMatrix(L1, size(M)) * M ≈ BandedBlockBandedMatrix(L1, size(M)) * M
  MethodError: no method matching view(::Tuple{Int64}, ::UnitRange{Int64})
  Closest candidates are:
    view(!Matched::AbstractArray, ::Any...) where N at subarray.jl:155
  Stacktrace:
   [1] Array(::DiffEqOperators.DerivativeOperator{Float64,1,false,Float64,StaticArrays.SArray{Tuple{3},Float64,1,3},StaticArrays.SArray{Tuple{0},StaticArrays.SArray{Tuple{3},Float64,1,3},1,0},Nothing,Nothing}, ::Tuple{Int64}) at /home/travis/build/JuliaDiffEq/DiffEqOperators.jl/src/derivative_operators/concretization.jl:232
   [2] top-level scope at /home/travis/build/JuliaDiffEq/DiffEqOperators.jl/test/concretization.jl:18
   [3] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
   [4] top-level scope at /home/travis/build/JuliaDiffEq/DiffEqOperators.jl/test/concretization.jl:12
```
I'll change this back. Shoulnd't the version without the view be exactly the same as the original version? Doesn't `for M_i in Mshape[1:N-1]`) create a copy? We can try to use a view regardless, but that wasn't my original intention.

Issue #181, opened 2019-09-21
Concretization of derivatives on non-uniform grids seems broken


jagot:
Trying to concretize derivatives on a non-uniform grid (which happens to be uniform...) always gives me the same 3x5 matrix, regardless of whether I use `Matrix`, `BandedMatrix`, `sparse`, &c:

```julia
julia> N = 10;

julia> dx = ones(Float64, N+1);

julia> L = CenteredDifference(2,2, dx, N);

julia> Matrix(L)
3×5 Array{Float64,2}:
 1.0  -2.0   1.0   0.0  0.0
 0.0   1.0  -2.0   1.0  0.0
 0.0   0.0   1.0  -2.0  1.0

julia> N = 50;

julia> dx = ones(Float64, N+1);

julia> L = CenteredDifference(2,2, dx, N);

julia> Matrix(L)
3×5 Array{Float64,2}:
 1.0  -2.0   1.0   0.0  0.0
 0.0   1.0  -2.0   1.0  0.0
 0.0   0.0   1.0  -2.0  1.0

julia> BandedMatrix(L)
3×5 BandedMatrix{Float64,Array{Float64,2},Base.OneTo{Int64}}:
 1.0  -2.0   1.0   0.0  0.0
 0.0   1.0  -2.0   1.0  0.0
 0.0   0.0   1.0  -2.0  1.0
```

Using explicitly uniform spacing works as intended:
```julia
julia> N = 50;

julia> dx = 1.0;

julia> L = CenteredDifference(2,2, dx, N);

julia> BandedMatrix(L)
50×52 BandedMatrix{Float64,Array{Float64,2},Base.OneTo{Int64}}:
 1.0  -2.0   1.0   0.0   0.0    ⋅     ⋅   …    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    ⋅
 0.0   1.0  -2.0   1.0   0.0   0.0    ⋅        ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    ⋅
 0.0   0.0   1.0  -2.0   1.0   0.0   0.0       ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    ⋅
 0.0   0.0   0.0   1.0  -2.0   1.0   0.0       ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    ⋅
[...]

Issue #226, opened 2020-02-18
Concretization struggles using CenteredDifference with nonuniform discretization spacing 


cristophermoen:
The concretization of the `DerivativeOperator` from the `CenteredDifference` function, i.e.,  `Array(CenteredDifference(derivative_order::Int,   approximation_order::Int, dx, len::Int))` does not seem to be working correctly when `dx` is an array describing a nonuniform discretization.

ChrisRackauckas:
Fixed!

Issue #141, opened 2019-07-17
CentredDifference problem at right boundary for odd derivative order and high approximation order


xtalax:
@ajozefiak @ChrisRackauckas @shivin9 The output from [this gist](https://gist.github.com/xtalax/872fc1dec342f91bb4a973528d74e5ed) which benchmarks the error for CenteredDifference at different derivative order and approximation order shows very high maximum error for odd derivative order and higher approximation order:
```
n = 10000
dor = 1
aor = 2
max_error = 6.495248771434689e-5
avg_error = 6.495239650094344e-5
dor = 1
aor = 3
max_error = 6.495248771434689e-5
avg_error = 6.495239650094344e-5
dor = 1
aor = 4
max_error = 200.00000000004513
avg_error = 0.020004000853293735
dor = 1
aor = 5
max_error = 199.99999999997743
avg_error = 0.02000400085327766
dor = 1
aor = 6
max_error = 199.99999999999827
avg_error = 0.04000800161049383
dor = 2
aor = 2
max_error = 3.2579010245858104e-5
avg_error = 3.247619589550224e-5
dor = 2
aor = 3
max_error = 1.2843858444852896e-7
avg_error = 1.0544439291065866e-8
dor = 2
aor = 4
max_error = 1.2843858444852896e-7
avg_error = 1.0532578093008475e-8
dor = 2
aor = 5
max_error = 1.3801313241358788e-7
avg_error = 1.1810916467400091e-8
dor = 2
aor = 6
max_error = 1.3801313241358788e-7
avg_error = 1.1811920672675504e-8
dor = 3
aor = 2
max_error = 199.99990895367293
avg_error = 0.020101412009188885
dor = 3
aor = 3
max_error = 199.9999797756774
avg_error = 0.020101410150574293
dor = 3
aor = 4
max_error = 200.00000851359022
avg_error = 0.04001363617120823
dor = 3
aor = 5
max_error = 200.00001516627336
avg_error = 0.040013636109185395
dor = 3
aor = 6
max_error = 200.0000123902663
avg_error = 0.060019264443696506
dor = 4
aor = 2
max_error = 0.08111202877485324
avg_error = 0.007821480500803916
dor = 4
aor = 3
max_error = 0.12526341730609303
avg_error = 0.012445415030591017
dor = 4
aor = 4
max_error = 0.12526341730609303
avg_error = 0.012454720240003657
dor = 4
aor = 5
[...]

PR #227, opened 2020-02-19
Irregular grid fix

fgerick:
I have fixed the irregular grid operator. The stencils should be constructed properly now. The concretization of the operator is fixed and I have removed the SVector of all stencil_coefs in the irregular grid case, as this is a very large array for large grids. I have added some tests and I found that there's maybe a need of finding a consistent formula for the error estimate of the finite difference operators as a function of dx, derivative order and approximation order (I guess it's not easy).


fgerick:
fixes #181 and #226  and #141 should also be fixed by this and my previous pr?

PR #228, opened 2020-02-28
Tests for non-uniform grids, Integrating GhostDerivativeOperator with AbstractDiffEqLinearOperator

chenwilliam77:
This pull request does . . .

1. Removes the use of `hasmethod` in the implementation of `*(coeff_func::Function, A::DerivativeOperator)` and requires the user to manually call `update_coefficients!` to actually compuet the coefficients.

2. Adds new tests for non-uniform grids, mainly in `bc_coeff_compositions.jl` and `robin.jl`.

3. Makes `GhostDerivativeOperator` a subtype of `AbstractDiffEqLinearOperator` so you can perform operations like `(L1 + L2) * Q` and `L1 *Q + L2 * Q`. Tests were also added to check that `(L1 + L2) * Q == L1 * Q + L2 * Q`


src/DiffEqOperators.jl
```diff
@@ -56,5 +60,6 @@ export DirichletBC, Dirichlet0BC, NeumannBC, Neumann0BC, RobinBC, GeneralBC, Mul
        MultiDimDirectionalBC, ComposedMultiDimBC,
        compose, decompose, perpsize
 export GhostDerivativeOperator
+export DiagonalOperator, IdentityOperator
```
ChrisRackauckas: These don't make a lot of sense. Isn't `DiagonalOperator` just an array operator with a `Diagonal` array? And `IdentitiyOperator`: does it do more than `I`?

src/composite_operators.jl
```diff
@@ -38,9 +38,9 @@ mul!(Y::AbstractVecOrMat, L::DiffEqScaledOperator, B::AbstractVecOrMat) =
 ldiv!(Y::AbstractVecOrMat, L::DiffEqScaledOperator, B::AbstractVecOrMat) =
   lmul!(1/L.coeff, ldiv!(Y, L.op, B))
 factorize(L::DiffEqScaledOperator) = L.coeff * factorize(L.op)
-for fact in (:lu, :lu!, :qr, :qr!, :cholesky, :cholesky!, :ldlt, :ldlt!,
```
ChrisRackauckas: please don't mix formatting changes in with other PRs.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -190,6 +185,19 @@ end
 function DiffEqBase.update_coefficients!(A::AbstractDerivativeOperator,u,p,t)
     if A.coeff_func !== nothing
         A.coeff_func(A.coefficients,u,p,t)
+    else
+        @warn "No coeff_func found. No updating performed."
```
ChrisRackauckas: it shouldn't print every time: the default is a pretty standard condition.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -190,6 +185,19 @@ end
...
+    end
```
ChrisRackauckas:
```suggestion
    end
```

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -190,6 +185,19 @@ end
...
+end
+function DiffEqBase.update_coefficients!(A::AbstractDerivativeOperator) where {T<:Number}
```
ChrisRackauckas: This one doesn't make sense: it's only in the API with `(u,p,t)`, and this wouldn't be widely usable.

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -128,6 +128,20 @@ function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
         A.high_boundary_coefs,coefficients,A.coeff_func)
 end
 
+function *(c::AbstractVector{<:Number}, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
+    coefficients = A.coefficients === nothing ? c .* ones(T, A.len) : c .* A.coefficients
```
ChrisRackauckas: supporting numbers using a `FillVector` would be good, because this would break isbits-ness.

test/runtests.jl
```diff
@@ -10,12 +10,13 @@ const is_TRAVIS = haskey(ENV,"TRAVIS")
 if GROUP == "All" || GROUP == "Interface"
     @time @safetestset "Utilities Tests" begin include("utils.jl") end
     @time @safetestset "Poisson example" begin include("../examples/poisson.jl") end
-    @time @safetestset "Heat equation example" begin include("../examples/heat_equation.jl") end
+#     @time @safetestset "Heat equation example" begin include("../examples/heat_equation.jl") end
```
ChrisRackauckas: why is this commented out? Tests should almost never be commented out and instead set as `@test_broken` as needed.

ChrisRackauckas:
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/operators/basic_operators.jl#L4

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/operators/basic_operators.jl#L87

I don't see a reason to have two versions of these.

src/DiffEqOperators.jl
chenwilliam77: My apologies, I didn't mean to make this part of the PR. Rewinding this branch (and getting rid of the DiagonalOperator and IdentityOperator).

src/composite_operators.jl
chenwilliam77: I'm not sure what you mean, but regardless, I don't recall intending to change any of these lines of code that are highlighted here.

src/derivative_operators/abstract_operator_functions.jl
chenwilliam77: I agree. I put there in case it maybe was needed, but I'm happy to delete these lines.

test/runtests.jl
chenwilliam77: I forgot to uncomment this test; apologies!

src/derivative_operators/derivative_operator_functions.jl
chenwilliam77: I meant for this to be a simple way of multiplying a `DerivativeOperator` by a vector of different coefficients. I'll be adding an assert statement at the beginning to check `c` is the right length and just passing `c` rather than `c .* ones(T, A.len)`.

src/derivative_operators/abstract_operator_functions.jl
```diff
@@ -186,19 +186,6 @@ end
 function DiffEqBase.update_coefficients!(A::AbstractDerivativeOperator,u,p,t)
     if A.coeff_func !== nothing
```
ChrisRackauckas:
```suggestion

```

src/derivative_operators/derivative_operator_functions.jl
```diff
@@ -115,7 +115,7 @@ function *(A::DerivativeOperator{T,N},M::AbstractArray{T}) where {T<:Real,N}
 end
 
 function *(c::Number, A::DerivativeOperator{T,N,Wind}) where {T,N,Wind}
-    coefficients = A.coefficients === nothing ? one(T) .* c : c .* A.coefficients
+    coefficients = A.coefficients === nothing ? c : c .* A.coefficients
```
ChrisRackauckas:
```suggestion
    coefficients = A.coefficients === nothing ? oneunit(T) .* c : c .* A.coefficients
```

I missed what you were doing here the first time.

ChrisRackauckas: Though if it defaulted to a FillVector we would remove an allocation. We should make sure an issue gets opened on this.

chenwilliam77: I believe we don't need to make the coefficient a vector, we can just convert it to type `T`. For example, run the following example (some version of which should be in the tests):
`L = CenteredDifference(1, 2, 1., 5)`
`c = 3.`
`x = rand(7)`
`cL = c * L`
`@assert length(cL.coefficients) == 1`
`cL * x == c * (L * x)`

ChrisRackauckas:
Awesome, great work!

PR #241, opened 2020-04-30
Enable 2nd order check

mindlike:
The "2nd_order_check" test set was working (after some small tweaks), so I re-enabled it. Changes are in this commit: https://github.com/SciML/DiffEqOperators.jl/commit/616e37846de3d6cc3673fb46353bc0e2d6f66f8b


ChrisRackauckas:
Looks good.

PR #246, opened 2020-06-06
Rebased #168

ChrisRackauckas:


ChrisRackauckas:
@grahamas @xtalax when you rebase you have to force push, not do a merge afterwards. So I cherrypicked one commit back and it's all good.

test/ghost_derivative_operators_interface.jl
```diff
@@ -7,7 +7,7 @@ using DiffEqOperators, Test
     A = L * Q
 
     @test ndims(A) == 2
-    @test size(A) == (10, 10)
-    @test size(A, 1) == 10
-    @test length(A) == 100
+    @test_broken size(A) == (10, 10)
```
ChrisRackauckas, line 10: how did these break?

test/3D_laplacian.jl
```diff
@@ -0,0 +1,36 @@
+using DiffEqOperators, DifferentialEquations, ProgressMeter
+function main()
```
ChrisRackauckas, line 2: @xtalax note that this isn't needed anymore now that we are using `@safetestsets`

ChrisRackauckas:
Not sure if this is perfect, but perfect might be the enemy of good here. Let's make sure this is passing tests and get it merged. Then we can continue to clean up the other little pieces. I think the next thing I want to do is to try and start getting a real documentation together for this library to really take stock of where we're at.

ChrisRackauckas:
It looks like there's a lot of failing tests? Are there commits that were not on that branch?

ChrisRackauckas:
Doesn't look like this can merge because there's a ton of broken tests.

test/ghost_derivative_operators_interface.jl
grahamas: one of my "rebase" commits changes `A.L.len` to `size(A.L, 2)` in the definition of `size(::GhostDerivativeOperator)` which all the broken tests rely on.  I don't see any reason whatsoever why I would have done that -- both master and xtalax have `A.L.len` and I'm aware that those aren't the same thing. So I can fix that, but where should I commit?

ChrisRackauckas: PR to this branch.

test/runtests.jl
```diff
@@ -33,6 +33,7 @@ if GROUP == "All" || GROUP == "Interface"
     @time @safetestset "Upwind Operator Interface" begin include("upwind_operators_interface.jl") end
     @time @safetestset "MOLFiniteDifference Interface" begin include("MOLtest.jl") end
     @time @safetestset "Basic SDO Examples" begin include("BasicSDOExamples.jl") end
+    @time @safetestset "3D laplacian Test" begin include("3D_laplacian.jl") end
```
grahamas, line 36: Something's weird -- I'm not sure this is from the right commit.  Line 23 of test/runtests.jl is causing an error that's fixed post-rebase (cf. https://github.com/xtalax/DiffEqOperators.jl/pull/26/files#diff-fce720c43af3c52c862fd7451c7374b8)

test/ghost_derivative_operators_interface.jl
grahamas: AH, my mistake.  Disregard my last comment -- I was, today, looking at xtalax:master rather than xtalax:#168 to figure out what I had been rebasing towards.  Luckily for my own personal sanity, I wasn't stupid enough to introduce my own ideas when I was rebasing.  Unfortunately, though, I'm not sure I can fix this broken test.  I marked it as broken because I saw that I could not fix it and I wanted to get past it to any tests I might be able to fix.  xtalax:#168 does use `size(A.L, 2)` in its definition of `size(::GhostDerivativeOperator)`.  That means that `size(A)` will return `(size(A.L, 2), size(A.L, 2))` which will return `(12,12)` in a very straightforward manner because in the constructor for CenteredDifference it sets the second dimension's size to be `10+2`.  I don't know which definition of `size` (or maybe the constructor?) is wrong.  I can start trying to fix the code's semantics, but that'll be a longer process.

ChrisRackauckas: I would revert that change I think?

grahamas:
That's my guess, too, but I don't know why @xtalax changed it in the first
place, or if other new code relies on the new definition.  If this PR is
still unresolved at the end of June (after a thesis committee meeting) I'd
be happy to jump in completely. Until then I wouldn't be comfortable
asserting a change like this. Still happy to resolve any rebase-related
problems in the meantime, since those are more mechanical and local. Though
depending on how long it takes to resolve the failing tests, I might just
need to run another rebase in a month....

ChrisRackauckas:
No worries. I think these test failures are what stalled it in the first place, so we can work through them whenever there's time.

test/3D_laplacian.jl
xtalax: noted, i'll change that around on my next commit

test/bc_coeff_compositions.jl
```diff
@@ -433,25 +445,57 @@ end
...
+    As_l, As_b = sparse(A, length(u))
 
-    # Check ldiv!
-    u_temp = zeros(N)
-    ldiv!(u_temp, A, u)
-    @test u_temp ≈ ghost_u ≈ analytic_u
+    @test As_l ≈ analytic_AL #This is true
+    @test As_b ≈  analytic_Ab #This is true
+    @test Vector(u .- analytic_Ab) ≈ Vector(u .- As_b) #this is true
+    @test_broken analytic_AL\Vector(u .- analytic_Ab) ≈ As_l\Vector(u .- As_b) #Then why isn't this working?
```
xtalax, line 467: @ChrisRackauckas  this error and those just below are what stalled progress before - I really have no idea why this isn't working

ChrisRackauckas: what's the error?

xtalax: the test marked broken should definitely pass if the above three are passing

test/ghost_derivative_operators_interface.jl
grahamas: Reverted and passing.  @ChrisRackauckas how often do you want me to PR from my personal branch into this branch?

test/bc_coeff_compositions.jl
grahamas: It feels like it has to be a bug in the one of the `\` implementations (or just imprecision, I guess). Running everything in the REPL:

```
julia> using DiffEqOperators

julia> dx = 0.01; x = 0.01:dx:0.2; N=length(x); u=sin.(x); L = CenteredDifference(4, 4, dx, N)
; Q = RobinBC((1.0, 0.0, 0.0), (1.0, 0.0, sin(0.2+dx)), dx); A = L*Q;                         
                                                                                              
julia> analytic_L = fourth_deriv_approx_stencil(N) ./ dx^4; analytic_QL = [transpose(zeros(N))
; Diagonal(ones(N)); transpose(zeros(N))]; analytic_AL = analytic_L * analytic_QL; analytic_Qb
 = [zeros(N+1); sin(0.2+dx)]; analytic_Ab = analytic_L*analytic_Qb; 

julia> analytic_AL \ Vector(u .- analytic_Ab) ≈ As_l \ Vector(u .- As_b)
false

julia> analytic_AL \ Vector(u .- analytic_Ab) ≈ Array(As_l) \ Vector(u .- As_b)
true

julia> methods(\, typeof.([analytic_AL, Vector(u .- analytic_Ab)]))
# 1 method for generic function "\":                                                                                         
[1] \(A::AbstractArray{T,2} where T, B::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T)
 in LinearAlgebra at /home/graham/.local/opt/julia-1.5.0/share/julia/stdlib/v1.5/LinearAlgebra
/src/generic.jl:1102

julia> methods(\, typeof.([As_l, Vector(u .- As_b)]))
# 1 method for generic function "\":
[1] \(A::SparseArrays.AbstractSparseMatrixCSC, B::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T) in SparseArrays at /home/graham/.local/opt/julia-1.5.0/share/julia/stdlib/v1.5/SparseArrays/src/linalg.jl:1460  
```

grahamas: I filed an issue with an MWE: https://github.com/JuliaLang/julia/issues/37004.  I'll mark this as broken for now and move on.

grahamas: They pointed out that the difference is within the expected range given the condition number of `analytic_AL`. I've revised the test to reflect that, and now it and a nearby similar test are passing. I imagine this is a problem in some other tests as well.  Would we rather use a better conditioned matrix?  I wouldn't know how to choose.

ChrisRackauckas: Just change the tolerance to be fit for the condition number. Any stencil operation is going to be pretty ill-conditioned (this is where the stiffness comes from)

grahamas:
Status update:  Only two problem areas remaining.

1. `upwind_operators_interface.jl` (test file) produces errors at random, varying between runs.  This obviously drives me insane, especially since I'm not even sure how the ghost derivative stuff should be affecting upwind operators, so I'm leaving it for last.

2. `3D_laplacian.jl` (test file) reveals that there's no `mul!` method for multidimensional operators -- they all assert that the padding is in only one dimension.  A cursory dive back into the pre-rebase commit history suggests I didn't accidentally delete anything.  That being the case, I'll have to rework the fallback in `derivative_operator_functions.jl` to do a check for higher dimensional padding.

grahamas:
The fallback `mul!` now works with multidimensionally padded inputs by introducing `unpadded_size` helper function. TODO:
1. Update specialized `mul!`s that call `NNlib.conv!`
2. Redefine `A*Q` as a composition so that `A*Q*u` just works (alternative is repadding `u` for every dimension... seems unwise)

ChrisRackauckas:
Awesome. Should this be passing then?

ChrisRackauckas:
Looks like there's a merge conflict.

grahamas:
@ChrisRackauckas Nope!  I don't have access to this branch; I was working on yet another fork.  I've opened #279 which is rebased to master as of 20 minutes ago and is passing all tests on my machine.

ChrisRackauckas:
alright, let's continue there.

src/composite_operators.jl
```diff
@@ -46,17 +51,19 @@ end
...
+    C<:AbstractVector{T}} <: AbstractDiffEqCompositeOperator{T}
+    ops::O
+    cache::C
+    function DiffEqOperatorCombination(ops; cache=nothing)
+        T = eltype(ops[1])
+        for i in 2:length(ops)
+            @assert size(ops[i]) == size(ops[1]) "Operators must be of the same size to be combined! Mismatch between $(ops[i]) and $(ops[i-1]), which are operators $i and $(i-1) respectively"
+        end
+        if cache == nothing
+            cache = Vector{T}(undef, size(ops[1], 1))
```
ChrisRackauckas, line 63: Let's zero this so errors with undef memory don't pop up.

PR #250, opened 2020-06-11
[WIP] First steps in PDE autodiscretization.

emmanuellujan:
### One-dimensional PDE auto-discretization using finite-difference (work in progress)
- Current PR is associated to issue: https://github.com/SciML/DifferentialEquations.jl/issues/469
- Auto-discretization code is located in src/MOL_discretization.jl. Tests are located in test/ directory.
- The following 1D CFD cases are supported, nonetheless more testing is needed.
    - Typical cases of the 1D diffusion equation:
        - [x] Dt(u(t,x)) ~ Dxx(u(t,x))
        - [x] Dt(u(t,x)) ~ D*Dxx(u(t,x))
        - [x] Dt(u(t,x)) ~ Dx(D(t,x))*Dx(u(t,x))+D(t,x)*Dxx(u(t,x))
    - Typical cases of the 1D convection equation:
        - [x] Dt(u(t,x)) ~ -Dx(u(t,x))
        - [x] Dt(u(t,x)) ~ -Dx(u(t,x)) + S, S is constant
        - [x] Dt(u(t,x)) ~ -v*Dx(u(t,x)), v is constant
        - [x] Dt(u(t,x)) ~ -Dx(v(t,x))*u(t,x)-v(t,x)*Dx(u(t,x)), v(t,x)=1
        - [x] Dt(u(t,x)) ~ -Dx(v(t,x))*u(t,x)-v(t,x)*Dx(u(t,x)), v(t,x)=0.999 + 0.001 * t * x 
    - Typical cases of 1D diffusion-convection equation:
        - [x] Dt(u(t,x)) ~ Dxx(u(t,x))-v*Dx(u(t,x))
    - Source/sink term
        - [x] All cases above support a constant source/sink term in rhs
        - [x] All cases above support a source/sink term which depends on (t,x) in rhs
    - Boundary conditions
        - [x] All cases above support Dirichlet boundary conditions

### DIFF
--- a/src/MOL_discretization.jl
+++ b/src/MOL_discretization.jl
@@ -49,7 +49,9 @@ function calc_coeff_mat(input,iv,grade,order,dx,m,nonderiv_depvars)
                     L = :([ (x=i*$dx;eval($expr)) for i=1:$m ])
                 end
             elseif grade == 1
-                L = :($(UpwindDifference(grade,order,dx,m,1.)))
+                # TODO: the discretization order should not be the same for
+                #       first derivatives and second derivarives
+                L = :($(UpwindDifference(grade,1,dx,m,1.)))
             else
                 L = :($(CenteredDifference(grade,order,dx,m)))
             end
@@ -57,14 +59,28 @@ function calc_coeff_mat(input,iv,grade,order,dx,m,nonderiv_depvars)
         elseif input.op isa Differential
             grade += 1
             return calc_coeff_mat(input.args[1],input.op.x,grade,order,dx,m,nonderiv_depvars)
-        elseif input.op isa typeof(-)
-            L = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
-            return Expr(:call,:*,:(-1),L)
         elseif input.op isa typeof(*)
-            #TODO: verificar si hay más de un argumento en input
             expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
             expr2 = calc_coeff_mat(input.args[2],iv,grade,order,dx,m,nonderiv_depvars)
             return Expr(:call,:*,expr1,expr2)
+        elseif input.op isa typeof(-)
+            if size(input.args,1) == 2
+                expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
+                expr2 = calc_coeff_mat(input.args[2],iv,grade,order,dx,m,nonderiv_depvars)
+                return Expr(:call,:-,expr1,expr2)
+            else #if size(input.args,1) == 1
+                expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
+                return Expr(:call,:*,:(-1),expr1)
+            end
+        elseif input.op isa typeof(+)
+            if size(input.args,1) == 2
+                expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
+                expr2 = calc_coeff_mat(input.args[2],iv,grade,order,dx,m,nonderiv_depvars)
+                return Expr(:call,:+,expr1,expr2)
+            else #if size(input.args,1) == 1
+                expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
+                return Expr(expr1)
+            end
         end
     end
 end
@@ -126,14 +142,19 @@ function DiffEqBase.discretize(pdesys::PDESystem,discretization::MOLFiniteDiffer
     u0 = @eval $g.($interior,$t)
     Q = DirichletBC(u_x0,u_x1)
 
+    # TODO: see TODO below related to "mul!(du,L,Q*u)"
+    Id = [ i==j ? 1 : 0 for i = 1:m+2, j = 1:m+2 ]
+
     ### Define the discretized PDE as an ODE function ##########################
     function f(du,u,p,t)
         for L in values(L_expr)
-            # TODO: there is probably a fancier way to eval time
             # TODO: is there a better way to use eval here?
             g = eval(:((t) -> $L))
             L = @eval $g.($t)
-            mul!(du,L,Q*u)
+
+            # TODO: mul!(du,L,Q*u) should work when L is a DiffEqOperatorCombination
+            #       L*Id should not be needed
+            mul!(du,L*Id,Q*u)
         end
     end
 
--- a/src/MOL_discretization.jl
+++ b/src/MOL_discretization.jl
@@ -40,17 +40,23 @@ end
 #            =>  Dx(u(t,x))=t*x*Dx(u(t,x))
 function calc_coeff_mat(input,iv,grade,order,dx,m,nonderiv_depvars)
     if input isa ModelingToolkit.Constant
-        return :($input.value)
+        x = abs(input.value)
+        return :($x)
     elseif input isa Operation
         if input.op isa Variable
             if haskey(nonderiv_depvars,input.op)
-                # TODO: Here, evaluate expression w.r.t space (x,y,z).
-                #       Then, in the "f" ODE function (in DiffEqBase.discretize),
-                #       evaluate expression w.r.t. time (t).
-                expr = Expr(nonderiv_depvars[input.op])
-                L = :(Diagonal([ (x=i*$dx;eval($expr)) for i=1:$m ]))
+                x = nonderiv_depvars[input.op]
+                if x isa ModelingToolkit.Constant
+                    L = :($x.value)
+                else
+                    # TODO: Here, evaluate expression w.r.t space (x,y,z).
+                    #       Then, in the "f" ODE function (in DiffEqBase.discretize),
+                    #       evaluate expression w.r.t. time (t).
+                    expr = Expr(x)
+                    L = :(Diagonal([ (x=i*$dx;eval($expr)) for i=1:$m ]))
+                end
             elseif grade == 1
-                L = :($(UpwindDifference(grade,order,dx,m,1)))
+                L = :($(UpwindDifference(grade,order,dx,m,-1)))
             else
                 L = :($(CenteredDifference(grade,order,dx,m)))
             end
@@ -60,7 +66,7 @@ function calc_coeff_mat(input,iv,grade,order,dx,m,nonderiv_depvars)
             return calc_coeff_mat(input.args[1],input.op.x,grade,order,dx,m,nonderiv_depvars)
         elseif input.op isa typeof(-)
             L = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
-            return Expr(:call, :*, :(-1), L)
+            return L
         elseif input.op isa typeof(*)
             # TODO: probably the code below can be shortened
             expr1 = calc_coeff_mat(input.args[1],iv,grade,order,dx,m,nonderiv_depvars)
--- a/src/MOL_discretization.jl
+++ b/src/MOL_discretization.jl
@@ -9,23 +9,45 @@ MOLFiniteDifference(args...;order=2) = MOLFiniteDifference(args,order)
 
 # Get boundary conditions from an array
 function get_bcs(bcs,tdomain,domain)
-    u_t0 = 0.0
-    u_x0 = 0.0
-    u_x1 = 0.0
+    lhs_deriv_depvars_bcs = Dict()
     n = size(bcs,1)
     for i = 1:n
-        if bcs[i].lhs.op isa Variable
-            if isequal(bcs[i].lhs.args[1],tdomain.lower) # u(t=0,x)
-                u_t0 = Expr(bcs[i].rhs)
+        var = bcs[i].lhs.op
+        if var isa Variable
+            if !haskey(lhs_deriv_depvars_bcs,var)
+                lhs_deriv_depvars_bcs[var] = Array{Expr}(undef,3)
+            end
+            if isequal(bcs[i].lhs.args[1],tdomain.lower) # u(t=0,x) 
+                lhs_deriv_depvars_bcs[var][1] = Expr(bcs[i].rhs)
             elseif isequal(bcs[i].lhs.args[2],domain.lower) # u(t,x=x_init)
-                u_x0 = bcs[i].rhs.value
+                lhs_deriv_depvars_bcs[var][2] = :(var=$(bcs[i].rhs.value))
             elseif isequal(bcs[i].lhs.args[2],domain.upper) # u(t,x=x_final)
-                u_x1 = bcs[i].rhs.value
+                lhs_deriv_depvars_bcs[var][3] = :(var=$(bcs[i].rhs.value))
             end
         end
     end
-    return (u_t0,u_x0,u_x1)
+    return lhs_deriv_depvars_bcs
 end
+
+# # Get boundary conditions from an array
+# function get_bcs(bcs,tdomain,domain)
+#     u_t0 = 0.0
+#     u_x0 = 0.0
+#     u_x1 = 0.0
+#     n = size(bcs,1)
+#     for i = 1:n
+#         if bcs[i].lhs.op isa Variable
+#             if isequal(bcs[i].lhs.args[1],tdomain.lower) # u(t=0,x)
+#                 u_t0 = Expr(bcs[i].rhs)
+#             elseif isequal(bcs[i].lhs.args[2],domain.lower) # u(t,x=x_init)
+#                 u_x0 = bcs[i].rhs.value
+#             elseif isequal(bcs[i].lhs.args[2],domain.upper) # u(t,x=x_final)
+#                 u_x1 = bcs[i].rhs.value
+#             end
+#         end
+#     end
+#     return (u_t0,u_x0,u_x1)
+# end
 
 # Recursively traverses the input expression (rhs), replacing derivatives by
 # finite difference schemes. It returns a time dependent expression (expr)
@@ -33,13 +55,13 @@ end
 # Note: 'non-derived' dependent variables are inserted into the diff. equations
 #       E.g. Dx(u(t,x))=v(t,x)*Dx(u(t,x)), v(t,x)=t*x
 #            =>  Dx(u(t,x))=t*x*Dx(u(t,x))
-function discretize_2(input,iv,grade,order,dx,m,nonderiv_depvars)
+function discretize_2(input,iv,grade,order,dx,m,lhs_nonderiv_depvars)
     if input isa ModelingToolkit.Constant
         return :($input.value)
     elseif input isa Operation
         if input.op isa Variable
-            if haskey(nonderiv_depvars,input.op)
-                x = nonderiv_depvars[input.op]
+            if haskey(lhs_nonderiv_depvars,input.op)
+                x = lhs_nonderiv_depvars[input.op]
                 if x isa ModelingToolkit.Constant
                     expr = :($x.value)
                 else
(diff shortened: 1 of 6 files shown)


test/MOL_1D_Linear_Convection.jl
```diff
@@ -0,0 +1,45 @@
+# 1D linear convection problem
```
ChrisRackauckas, line 1: These need to be added to runtests.jl to run.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,81 @@
...
+    X = domain.lower:dx:domain.upper
+    m = size(X)[1]-2
+    L = calc_coeff_mat(pdesys.eq.rhs,pdesys.indvars[2],0,discretization.order,dx,m)
+    (u_t0,u_x0,u_x1) = extract_bc(pdesys.bcs,tdomain,domain)
+    Q = DirichletBC(u_x0,u_x1)
+    function f(du,u,p,t)
+        mul!(du,L,Array(Q*u))
+    end
+    t = 0.0
+    u0 = eval_expr(u_t0,interior,t)
```
ChrisRackauckas, line 77: we can merge without fixing this part, but let's get an issue open about not using eval here. @shashi probably has a solution.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,81 @@
...
+    len = domain.upper-domain.lower
+    dx = discretization.dxs[1]
+    interior = domain.lower+dx:dx:domain.upper-dx
+    X = domain.lower:dx:domain.upper
+    m = size(X)[1]-2
+    L = calc_coeff_mat(pdesys.eq.rhs,pdesys.indvars[2],0,discretization.order,dx,m)
+    (u_t0,u_x0,u_x1) = extract_bc(pdesys.bcs,tdomain,domain)
+    Q = DirichletBC(u_x0,u_x1)
+    function f(du,u,p,t)
+        mul!(du,L,Array(Q*u))
```
ChrisRackauckas, line 74:
```suggestion
        mul!(du,L,Q*u)

```
you don't need to allocate the array here. This will use a lazy concatenation to decrease the memory cost.

ChrisRackauckas:
Looking great. @YingboMa comments?

test/MOL_1D_Linear_Convection.jl
emmanuellujan: Done.

Note: there are other tests that are currently failing (at least on my computer) when the file runtests.jl is executed. E.g. DiffEqOperators.jl/test/concretization.jl:
First Dimension: Error During Test at .../DiffEqOperators.jl/test/concretization.jl:18
  Test threw exception
  Expression: L1 * M ≈ Array(L1, size(M)) * vec(M) ≈ sparse(L1, size(M)) * M ≈ BandedMatrix(L1, size(M)) * M ≈ BandedBlockBandedMatrix(L1, size(M)) * M
...

src/MOL_discretization.jl
emmanuellujan: I will solve it before merge :)

ChrisRackauckas: I wouldn't worry about that. Just get the low hanging fruit, merge, and next PR. Shorter and quicker PRs are much easier to review IMO.

ChrisRackauckas:
I'll look into the unrelated test failures on master.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+   f = eval(:((x,y) -> $expr))
+   return @eval $f.($x,$y)
+end
+
+# Obtain boundary conditions from an array
+function extract_bc(bcs,tdomain,domain)
+    u_t0 = 0.0
+    u_x0 = 0.0
+    u_x1 = 0.0
+    n = size(bcs)[1]
```
YingboMa, line 19:
```suggestion
    n = size(bcs, 1)
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+    for i = 1:n
+        if isa(bcs[i].lhs.op,Variable)
```
YingboMa, line 21:
```suggestion
        if bcs[i].lhs.op isa Variable
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+   return @eval $f.($x,$y)
+end
+
+# Obtain boundary conditions from an array
+function extract_bc(bcs,tdomain,domain)
+    u_t0 = 0.0
+    u_x0 = 0.0
+    u_x1 = 0.0
+    n = size(bcs)[1]
+    for i = 1:n
```
YingboMa, line 20: I don't know why is there a loop if we only return scalars.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+                u_x1 = bcs[i].rhs.value
+            end
+        end
+    end
+    return (u_t0,u_x0,u_x1)
+end
+
+# Calculate coefficient matrix of the finite-difference scheme
+function calc_coeff_mat(input,iv,grade,order,dx,m)
+    if isa(input,ModelingToolkit.Constant)
```
YingboMa, line 36:
```suggestion
    if input isa Constant
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+            return input.value
+    elseif isa(input,Operation)
```
YingboMa, line 38:
```suggestion
    elseif input isa Operation
```

ChrisRackauckas:
Retriggering build due to upstream fixes

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
+            return L
+        elseif isa(input.op,Differential)
+            grade += 1
+            calc_coeff_mat(input.args[1],input.op.x,grade,order,dx,m)
+        elseif isa(input.op,typeof(*))
+            n = size(input.args)[1]
+            output = calc_coeff_mat(input.args[1],iv,grade,order,dx,m) 
+            for i = 2:n
+                output *= calc_coeff_mat(input.args[i],iv,grade,order,dx,m) 
+            end
```
YingboMa, line 54:
```suggestion
            output = prod(i -> calc_coeff_mat(input.args[i],iv,grade,order,dx,m), eachindex(input.args))
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
-  u0 = @. - interior * (interior - 1) * sin(interior)
-  PDEProblem(ODEProblem(f,u0,(tdomain.lower,tdomain.upper),nothing),Q,X)
+    tdomain = pdesys.domain[1].domain
+    domain = pdesys.domain[2].domain
+    @assert domain isa IntervalDomain
+    len = domain.upper-domain.lower
+    dx = discretization.dxs[1]
+    interior = domain.lower+dx:dx:domain.upper-dx
+    X = domain.lower:dx:domain.upper
+    m = size(X)[1]-2
```
YingboMa, line 69:
```suggestion
    m = size(X, 1)-2
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,80 @@
...
-  function f(du,u,p,t)
-    mul!(du,L,Array(Q*u))
-  end
-  u0 = @. - interior * (interior - 1) * sin(interior)
-  PDEProblem(ODEProblem(f,u0,(tdomain.lower,tdomain.upper),nothing),Q,X)
+    tdomain = pdesys.domain[1].domain
+    domain = pdesys.domain[2].domain
+    @assert domain isa IntervalDomain
+    len = domain.upper-domain.lower
+    dx = discretization.dxs[1]
```
YingboMa, line 66: Do we want to check `all(isequal(first(discretization.dxs)), discretization.dxs)`?

ChrisRackauckas:
Rebase since master now passes tests

src/MOL_discretization.jl
emmanuellujan: Okay, I opened a new issue: https://github.com/SciML/DiffEqOperators.jl/issues/254

src/MOL_discretization.jl
emmanuellujan: We return scalars or expressions (see u_t0 = Expr(bcs[i].rhs). Also, this function will increase its complexity by adding Neumann/Robin BCs and 2D/3D BCs.

We need the loop because the BCs can be presented in any order, e.g.:
```
bcs = [u(0,x) ~ -x*(x-1)*sin(x),
       u(t,0) ~ 0.0,
       u(t,1) ~ 0.0]
```
or
```
bcs = [ u(t,0) ~ 0.0,
        u(t,1) ~ 0.0,
        u(0,x) ~ -x*(x-1)*sin(x)]
```
etc.

src/MOL_discretization.jl
emmanuellujan: We will work with non-uniform mesh in other PRs :)

test/MOL_1D_Linear_Diffusion.jl
```diff
@@ -0,0 +1,198 @@
...
+    using OrdinaryDiffEq
+    sol = solve(prob,Tsit5(),saveat=0.1)
+
+    # Plot and save results
+    using Plots
+    plot(prob.space,Array(prob.extrapolation[1]*sol[:,1,1]))
+    plot!(prob.space,Array(prob.extrapolation[1]*sol[:,1,2]))
+    plot!(prob.space,Array(prob.extrapolation[1]*sol[:,1,3]))
+    plot!(prob.space,Array(prob.extrapolation[1]*sol[:,1,4]))
+    savefig("MOL_1D_Linear_Diffusion_Test01.png")
```
ChrisRackauckas, line 96: shouldn't save in tests

src/MOL_discretization.jl
```diff
@@ -1,22 +1,224 @@
...
+
+    end
+
+
+    Qu = Array{Float64}(undef,xx[1]+2,length(discretization))
+
+    ### Define the discretized PDE as an ODE function ##########################
+    function f(du,u,p,t)
+        for j in 1:length(discretization)
+            Qu[:,j] = Q[j]*u[:,j]
```
ChrisRackauckas, line 210: views

src/MOL_discretization.jl
```diff
@@ -1,22 +1,224 @@
...
+    ### Define the discretized PDE as an ODE function ##########################
+    function f(du,u,p,t)
+        for j in 1:length(discretization)
+            Qu[:,j] = Q[j]*u[:,j]            
+        end
+        j = 1
+        for (var,disc) in discretization
+            g = eval(:((u,t,i) -> $disc))
+            for i = 1:xx[1]
+                du[i,j] = @eval $g($(Qu),$t,$(i+1))
```
ChrisRackauckas, line 216: eval in the function will make it slow. The functions should be generated first and then enclosed.

ChrisRackauckas:
It's looking quite good, but this is too much in one PR. Could this get to a point of completion to merge and have other things then added? Gigantic PRs that do everything in one go are never a good thing.

src/MOL_discretization.jl
emmanuellujan: I replaced it with the code below, but I am not sure if this is how it should be done.
```Qu = [view(Q[j]*u[:,j],:) for j=1:length(discretization)]```

File: test/MOL_1D_Linear_Diffusion.jl (hunk 12 above)
Comment (emmanuellujan) [reply test/MOL_1D_Linear_Diffusion.jl:96]: Fixed

File: src/MOL_discretization.jl (hunk 14 above)
Comment (emmanuellujan) [reply src/MOL_discretization.jl:216]: I replaced it with a code analogous to the code below, but I am not sure if this is how it should be done.
```
expr = discretize(...)
discretization[var] = @eval (u,t,i) -> $expr
...
function f(du,u,p,t)
...
... = Base.invokelatest(discretization[var],Qu,t,i+1)
...
end
```

**Comment by emmanuellujan** [comment]
@ChrisRackauckas I will reduce the objectives of the PR to the implementation of one last test: a Saint Venant problem. It should be enough with what is already implemented: 1d, non-linear, coupled system, dirichlet bcs.

**Comment by ChrisRackauckas** [comment]
It would be good to get something merged here. I think some of the internals need to converge with @KirillZubov 's PINN extensions https://nextjournal.com/kirill_zubov/physics-informed-neural-networks-pinns-solver-on-julia-gsoc-2020-final-report so this project is falling a bit behind. Also, it needs to start taking into account the boundary changes coming in https://github.com/SciML/ModelingToolkit.jl/issues/526

**Comment by emmanuellujan** [comment]
@ChrisRackauckas Okay, I marked the PR as ready for review :-)

File: src/MOL_discretization.jl [hunk 15]
```diff
@@ -1,22 +1,220 @@
...
+                expr1 = discretize_2(input.args[1],grade,order,dx,lhs_deriv_depvars,lhs_nonderiv_depvars)
+                expr2 = discretize_2(input.args[2],grade,order,dx,lhs_deriv_depvars,lhs_nonderiv_depvars)
+                return Expr(:call,:+,expr1,expr2)
+            else #if size(input.args,1) == 1
+                expr1 = discretize_2(input.args[1],grade,order,dx,lhs_deriv_depvars,lhs_nonderiv_depvars)
+                return Expr(expr1)
+            end
+        end
+    end
+end
```
ChrisRackauckas, line 96: All of this code is quite unnecessary. It should just be a call to `CenteredDifference` and `UpwindDifference`

src/MOL_discretization.jl
```diff
@@ -1,22 +1,220 @@
...
+                lhs_deriv_depvars_bcs[var][1] = Expr(bcs[i].rhs)
+            elseif isequal(bcs[i].lhs.args[2],domain.lower) # u(t,x=x_init)
+                lhs_deriv_depvars_bcs[var][2] = :(var=$(bcs[i].rhs.value))
+            elseif isequal(bcs[i].lhs.args[2],domain.upper) # u(t,x=x_final)
+                lhs_deriv_depvars_bcs[var][3] = :(var=$(bcs[i].rhs.value))
+            end
+        end
+    end
+    return lhs_deriv_depvars_bcs
+end
```
ChrisRackauckas, line 30: This should be building a Q operator?

ChrisRackauckas:
It looks like tests are failing?

ChrisRackauckas:
Let's have a call to get in sync.

src/MOL_discretization.jl
emmanuellujan: The aim of this algorithm is to traduce an rhs such as: 
```
ln( v(t,x)^Dx(phi(t,x)) / (u(t,x)*(1-u(t,x))) )
where, v(t,x) = 1+2*t*x
```
to an expression like this:
```
ln(  (1+2*t*i*0.025)^((u[1][i]-u[1][i-1])/0.025) / (u[1][i]*(1-u[1][i])) )
```
where x=i * dx=i * 0.025, and "u" stores each variable (phi has index 1). 

The obtained expression is then evaluated at each time step in the function "f(du,u,p,t)". Using this strategy I avoid using matrices, which is convenient in non-linear problems, and I store every variable within "u", which is consistent with the "f" interface. How would you translate the rhs that I gave using CenteredDifference?


On the other hand, last month I have been thinking how to reimplement this prototype code in a simpler way. Here is an alternative:

Assuming lhs has expressions of this form: Dx(phi(t,x)) or v(t,x)

For each rhs (e.g. Dx(v(t,x)*phi(t,x))) associated to an lhs variable (e.g. phi)
1. expand derivatives (does mtk provide this feature?)
```
     e.g.: Dx(v(t,x)*phi(t,x))
             => Dx(v(t,x))*phi(t,x)+v(t,x)*Dx(phi(t,x))
```
2. replace derivatives and variables with its discretizations
```
     e.g.: Dx(v(t,x))*phi(t,x)+v(t,x)*Dx(phi(t,x))
            => ((u[1][i]-u[1][i-1])/0.025)*u[2][i]+u[1][i]*((u[2][i]-u[2][i-1])/0.025)
```
3. simplify obtained expression (does mtk provide this feature?)
4. solve the ODE problem

src/MOL_discretization.jl
emmanuellujan: It is a possibility, we should take into account that the BCs should be evaluated w.r.t. t at each time step within "f(du,u,p,t)".

emmanuellujan:
> Let's have a call to get in sync.

I download the github code, at least in my notebook, the tests associated with this code are working fine.

ChrisRackauckas:
>I download the github code, at least in my notebook, the tests associated with this code are working fine.

https://travis-ci.com/github/SciML/DiffEqOperators.jl/builds/180792060#L602-L613

The failure is in the new code so I think it must be related.

src/MOL_discretization.jl
ChrisRackauckas: Yes, Q is lazy so you'd just create the type on the fly before evaluation.

src/MOL_discretization.jl
ChrisRackauckas: >to an expression like this:

You don't want the expression though, since with that form you cannot hit the stencil compiler. This is the reason for the operator form since otherwise you will not have speed or GPU-compatibility.

>How would you translate the rhs that I gave using CenteredDifference?

That's an upwind operator call and then a broadcast expression.

emmanuellujan: > You don't want the expression though, since with that form you cannot hit the stencil compiler. This is the reason for the operator form since otherwise you will not have speed or GPU-compatibility.

The whole code is based on building the discretizations as expressions that can be evaluated within f(du,u,p,t), avoiding expressions implies reimplementing many things.

> That's an upwind operator call and then a broadcast expression.

Could you please explicitly write the outcome you expect to get when an Operation like the one below is discretized?:
```
ln( v(t,x)^Dx(phi(t,x)) / (u(t,x)*(1-u(t,x))) )
```

src/MOL_discretization.jl
emmanuellujan: Ok. I assume this is planned for an upcoming PR specializing in BCs, is that right?

ChrisRackauckas: That should already work

src/MOL_discretization.jl
ChrisRackauckas: `Dx(phi(t,x))` would be a forward difference operator, and the rest would be a broadcast expression

emmanuellujan: > Dx(phi(t,x)) would be a forward difference operator, 

That I had understood.

> and the rest would be a broadcast expression

Here is where I do not follow you, would you please write a single line of code with the discretized rhs, using a broadcast expression and the UpwindDifference operator?

```
rhs = ln( v(t,x)^Dx(phi(t,x)) / (u(t,x)*(1-u(t,x))) )
discretized_rhs = ...
```

This code should discretize the derivative, with UpwindDifference, but also the other variables: v and u. Furthermore, this discretized code (discretized_rhs) should be evaluated w.r.t. t at each time iteration, that is one of the reasons I used expressions to build the discretization.

src/MOL_discretization.jl
emmanuellujan: Note: the function get_bcs works, the Q operator is created below in the code:
```
### Get boundary conditions ################################
    # TODO: generalize to N equations
    lhs_deriv_depvars_bcs = get_bcs(pdesys.bcs,tdomain,domain[1])
    t = 0.0
    interior = domain[1].lower+dx[1]:dx[1]:domain[1].upper-dx[1]
    u0 = Array{Float64}(undef,length(interior),length(discretization))
    Q = Array{RobinBC}(undef,length(discretization))
    
    i = 1
    for var in keys(discretization)
        bcs = lhs_deriv_depvars_bcs[var]
        g = eval(:((x,t) -> $(bcs[1])))
        u0[:,i] = @eval $g.($interior,$t)
        u_x0 = eval(bcs[2])
        u_x1 = eval(bcs[3])
        Q[i] = DirichletBC(u_x0,u_x1)
        i = i+1
    end
```

emmanuellujan:
> The failure is in the new code so I think it must be related.

This is the method of lines discretization scheme

```
struct MOLFiniteDifference{T} <: DiffEqBase.AbstractDiscretization
  dxs::T
  order::Int
end
MOLFiniteDifference(args...;order=2) = MOLFiniteDifference(args,order)
```

If you invoke MOLFiniteDifference this way:
```
MOLFiniteDifference(0.1) 
or 
dx = 0.1
MOLFiniteDifference(dx) 
```
Then ```typeof(dxs)``` is a ```Tuple{Float64}```, and ```dxs=(0.1,)```

But if you invoke MOLFiniteDifference this way:
```
dx = 0.1
order = 2 
MOLFiniteDifference(dx,order) 
```
Then ```typeof(dxs)``` is a ```Float64```

This produced the error when parsing dxs. Any idea about how to modify MOLFiniteDifference to avoid the Tuple case?

ChrisRackauckas:
Oh, that needs to be an inner constructor then:

```julia
struct MOLFiniteDifference{T} <: DiffEqBase.AbstractDiscretization
  dxs::T
  order::Int
  MOLFiniteDifference(args...;order=2) = new{typeof(args)}(args,order)
end
```

emmanuellujan:
> Oh, that needs to be an inner constructor then:

This way makes dxs=(0.1,2) and order=2. I added a small change so dxs is not a tuple:

```
struct MOLFiniteDifference{T} <: DiffEqBase.AbstractDiscretization
    dxs::T
    order::Int
    MOLFiniteDifference(args...;order=2) = new{typeof(args[1])}(args[1],order)
end
```

src/MOL_discretization.jl
```diff
@@ -1,22 +1,236 @@
...
+                j = dep_var_idx[var]
+                if deriv_order == 0
+                    expr = :(u[:,$j])
+                elseif deriv_order == 1
+                    # TODO: approx_order and forward/backward should be 
+                    #       input parameters of each derivative
+                    approx_order = 1
+                    L = UpwindDifference(deriv_order,approx_order,dx[i],len[i]-2,-1)
+                    expr = :(-1*($L*Q[$j]*u[:,$j]))
+                elseif deriv_order == 2
```
ChrisRackauckas, line 72: The difference here isn't 1 or 2, but more generally, when it's odd you want to use upwinding and when it's even you want to use centered difference.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,236 @@
...
+    #                E.g. du/dt = d2u/dx2 + d2u/dy2 + f(t,x,y)
+    #        1.b) Stationary
+    #                There is more than one indep. variable, 't' is not included
+    #                E.g. 0 = d2u/dx2 + d2u/dy2 + f(x,y)
+    #   2) ODE System
+    #        't' is the only independent variable
+    #        The ODESystem is packed inside a PDESystem
+    #        E.g. du/dt = f(t)
+    #
+    #   Note: regarding input format, lhs must be "du/dt" or "0".
```
ChrisRackauckas, line 117: Note that it shouldn't always be an ODEProblem. If there isn't a "preferred dimension", then it should become a NonlinearProblem. We should discuss this a bit more.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,236 @@
...
+                i = indep_var_idx[deriv_var[1]]
+                j = dep_var_idx[var]
+                if deriv_order == 0
+                    expr = :(u[:,$j])
+                elseif deriv_order == 1
+                    # TODO: approx_order and forward/backward should be 
+                    #       input parameters of each derivative
+                    approx_order = 1
+                    L = UpwindDifference(deriv_order,approx_order,dx[i],len[i]-2,-1)
+                    expr = :(-1*($L*Q[$j]*u[:,$j]))
```
ChrisRackauckas, line 71: yup that's what I had in mind, basically. Though the way you're writing it is going to allocate and it's not going to extend to higher dimensions. We should talk next about how to get to higher dimensions.

src/MOL_discretization.jl
```diff
@@ -1,22 +1,236 @@
...
+    u_x0 = Array{Any}(undef,no_dep_vars)
+    u_x1 = Array{Any}(undef,no_dep_vars)
+    Q = Array{RobinBC}(undef,no_dep_vars)
+    
+    for var in keys(dep_var_idx)
+        j = dep_var_idx[var]
+        bcs = lhs_deriv_depvars_bcs[var]
+
+        g = eval(:((x,t) -> $(bcs[1])))
+        u_t0[:,j] = @eval $g.($(X[2][2:len[2]-1]),$t)
```
ChrisRackauckas, line 200: This should make use of https://github.com/SciML/RuntimeGeneratedFunctions.jl

ChrisRackauckas:
This looks good to merge. I made a few comments on things I saw that need to change, but they can come in follow ups and some of them are very simple. Thanks!

ChrisRackauckas:
🎉

emmanuellujan:
Great!

Thanks for the comments, I will take them into account for the next PR :-)

ChrisRackauckas:
I opened an issue per comment to keep track of them.

emmanuellujan:
> I opened an issue per comment to keep track of them.

Excellent

Issue #237, opened 2020-04-14
Support Complex Numbers?


AndiMD:
Can we allow complex numbers? Example:
```
RobinBC( (0.5+0im,0.5*im, 1.0+0im), (0.5+0im,0.5*im, 0.0+0im), 1.0)

ERROR: LoadError: MethodError: no method matching RobinBC(::Tuple{Complex{Float64},Complex{Float64},Complex{Float64}}, ::Tuple{Complex{Float64},Complex{Float64},Complex{Float64}}, ::Float64)
```

ChrisRackauckas:
Yes, we should probably just change the type signatures to allow this (and test it of course)

AndiMD:
this will also allow optimizing boundaries with `ForwardDiff.Dual`, which appears to be used by certain ODE solvers.

ChrisRackauckas:
Yes, I would not be opposed to seeing this change.

ChrisRackauckas:
It should probably just take changing the type signatures all to `<:Number`

volkerkarle:
Hi folks, may I ask what is the progress for this and if there is something to do? I wonder what is the technical current state, can we implement complex PDEs such as Gross–Pitaevskii equations?
Thanks!

PR #259, opened 2020-08-08
[WIP] (issue #237) Allow complex etc. function values

AndiMD:
Hi there,

This should allow inner constructors for RobinBC and GeneralBC to take different types between function values and grid spacing (#237).

- At the moment, there are no restrictions on the types, since the surrounding code is pretty unrestricted.
- The dot product `a⋅b` complex conjugates `a`, so I replaced it with `a' ⋅b`. Is there a better way?
- RobinBC is tested with ComplexF64, GeneralBC is not tested yet

Suggestions welcome! I'm still new to the Github workflow, so please be gentle ;)


AndiMD:
this did not quite work as intended - will have to work on this some more

PR #279, opened 2020-09-27
Multidimensional handling for GhostDerivativeOperators (part 3)

### LINKED ISSUE #168: Missing Method + for GhostDerivativeOperator

grahamas:
Bugfixed and rebased version of #246, which was a rebase of #170, which fixed #168.

Vast majority of code written by @xtalax.


grahamas:
Notes for reviewer:  Most of the work I've done since the last review was in `mul!` methods.  However, I didn't touch the `mul!` methods specialized for 2/3D and *composite* operators:  it seems to me that these fall back on the `mul!` operators that I *did* fix, and so they should also work.

Project.toml
```diff
@@ -38,6 +38,7 @@ SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 SimpleDifferentialOperators = "7ccca996-40d6-5c2f-be7f-4eb2934af9c4"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
ChrisRackauckas, line 41: tests shouldn't depend on DifferentialEquations instead use the components like OrdinaryDiffEq directly.

src/DiffEqOperators.jl
```diff
@@ -53,8 +56,9 @@ export AnalyticalJacVecOperator, JacVecOperator, getops
 export AbstractDerivativeOperator, DerivativeOperator,
        CenteredDifference, UpwindDifference
 export DirichletBC, Dirichlet0BC, NeumannBC, Neumann0BC, RobinBC, GeneralBC, MultiDimBC, PeriodicBC,
-       MultiDimDirectionalBC, ComposedMultiDimBC,
-       compose, decompose, perpsize
+       MultiDimDirectionalBC, ComposedMultiDimBC
+export compose, decompose, perpsize, c2l
```
ChrisRackauckas, line 60: we shouldn't export a short name like `c21`. I'll see what that's all about further down I assume

src/composite_operators.jl
```diff
@@ -27,15 +28,19 @@ opnorm(L::DiffEqScaledOperator, p::Real=2) = abs(L.coeff) * opnorm(L.op, p)
...
-ldiv!(Y::AbstractVecOrMat, L::DiffEqScaledOperator, B::AbstractVecOrMat) =
+*(L::DiffEqScaledOperator, x::AbstractArray) = L.coeff * (L.op * x)
+*(x::AbstractArray, L::DiffEqScaledOperator) = (L.op * x) * L.coeff
+/(L::DiffEqScaledOperator, x::AbstractArray) = L.coeff * (L.op / x)
+/(x::AbstractArray, L::DiffEqScaledOperator) = 1/L.coeff * (x / L.op)
+\(L::DiffEqScaledOperator, x::AbstractArray) = 1/L.coeff * (L.op \ x)
+\(x::AbstractArray, L::DiffEqScaledOperator) = L.coeff * (x \ L)
+for N in (2,3)
+  @eval begin
+    mul!(Y::AbstractArray{T,$N}, L::DiffEqScaledOperator{T}, B::AbstractArray{T,$N}) where {T} =
```
ChrisRackauckas, line 39: why does this need to depend on `N`? Can't it be generic? Or is this to fix some kind of ambiguity issues?

Project.toml
grahamas: No tests even use OrdinaryDiffEq anymore; I'll remove it now.

ChrisRackauckas: There should be some integration tests?

src/derivative_operators/bc_operators.jl
```diff
@@ -57,8 +80,11 @@ struct RobinBC{T, V<:AbstractVector{T}} <: AffineBC{T}
     end
 end
 
+stencil(q::AffineBC{T}, N::Int) where T = ([transpose(q.a_l) transpose(zeros(T, N-length(q.a_l)))], [transpose(zeros(T, N-length(q.a_r))) transpose(q.a_r)])
+affine(q::AffineBC) = (q.b_l, q.b_r)
 
-
+stencil(q::PeriodicBC{T}, N::Int) where T= ([transpose(zeros(T, N-1)) one(T)], [one(T) transpose(zeros(T, N-1))])
+affine(q::PeriodicBC{T}) where T = (zero(T), zero(T))
```
ChrisRackauckas, line 87: I don't understand this one.

src/derivative_operators/concretization.jl
```diff
@@ -163,57 +163,194 @@ function Base.convert(::Type{AbstractMatrix},A::AbstractBC{T}) where T
...
+    end
+    for dim in 1:N #Loop over boundaries
+        r_ = deepcopy(ranges)
+        r_[dim] = 1
+        lower = CartesianIndices((Tuple(r_))) #set up upper anmd lower indices
+        r_[dim] = s_pad[dim]
+        upper = CartesianIndices((Tuple(r_)))
+        for K in CartesianIndices(upper) #for every element of the boundaries
+            I = CartesianIndex(Tuple(K)[setdiff(1:N, dim)]) #convert K to 2D index for indexing the BC arrays
+            il = c2l(lower[K], s_pad) #Translate to linear indices
```
ChrisRackauckas, line 242: c21 looks like it could have a better name, and why would a user want to use this? It seems internal.

Project.toml
grahamas: Whoops yes.  Just removing `DifferentialEquations`, not `OrdinaryDiffEq`.  Meant to say that the test that formerly used `DifferentialEquations` now doesn't use anything of the sort.

src/derivative_operators/ghost_derivative_operator.jl
```diff
@@ -1,74 +1,39 @@
...
 end
 
 function *(L::AbstractDiffEqCompositeOperator{T}, Q::AbstractBC{T}) where{T}
     return sum(map(op -> op * Q, L.ops))
 end
 
-function LinearAlgebra.mul!(x::AbstractVector, A::GhostDerivativeOperator, u::AbstractVector)
-    @assert length(u) == A.L.len == length(x)
+function *(A::GhostDerivativeOperator{T1}, u::AbstractArray{T2}) where {T1,T2}
+    #TODO Implement a function domaincheck(L::AbstractDiffEqLinearOperator, u) to see if components of L along each dimension match the size of u
```
ChrisRackauckas, line 15: that would be nice. Worth opening an issue.

src/composite_operators.jl
grahamas: Based on the commit comment ("trying to make julia call the right methods")  I would guess the latter. But this comment and below are more for @xtalax.

src/derivative_operators/ghost_derivative_operator.jl
```diff
@@ -1,74 +1,39 @@
...
-    LinearAlgebra.mul!(M_temp, A, M)
-    return M_temp
-end
-
-
-function LinearAlgebra.ldiv!(x::AbstractVector, A::GhostDerivativeOperator, u::AbstractVector)
-    @assert length(x) == A.L.len
-    (AL,Ab) = Array(A)
-    LinearAlgebra.ldiv!(x, lu!(AL), u-Ab)
-end
```
ChrisRackauckas, line 45: Where did these dispatches go?

src/derivative_operators/multi_dim_bc_operators.jl
```diff
@@ -51,7 +51,7 @@ struct MultiDimDirectionalBC{T<:Number, B<:AtomicBC{T}, D, N, M} <: MultiDimensi
 end
 
 struct ComposedMultiDimBC{T, B<:AtomicBC{T}, N,M} <: MultiDimensionalBC{T, N}
-    BCs::Vector{Array{B, M}}
+    BCs::Vector{Array{B, M}} # Why isn't this a vector of MultiDimBCs?
```
ChrisRackauckas, line 54: Looks like it's a vector of `AtomicBC`. I'll keep reading to see if that makes sense.

ChrisRackauckas: Yeah seems like that works.

src/utils.jl
```diff
@@ -13,6 +13,14 @@ function unit_indices(N::Int) #create unit CartesianIndex for each dimension
     Tuple(out)
 end
 
+function c2l(I::CartesianIndex, s)  #Not sure if there is a builtin that does this - convert cartesian index to linear index of an array of size s
```
ChrisRackauckas, line 16: yeah this doesn't need to be exported.

src/derivative_operators/ghost_derivative_operator.jl
grahamas: I'll open an issue.  I basically wrote the code for this in one of the `mul!' methods, but with a caveat.

test/bc_coeff_compositions.jl
```diff
@@ -433,25 +445,63 @@ end
...
-
-    # Check ldiv! for Matrix
-    M_temp = zeros(N,3)
-    ldiv!(M_temp, A, M2)
-    @test M_temp ≈ ghost_M ≈ analytic_M
+    s = size(M2)
+    Qx = MultiDimBC{1}(Q, size(M2))
+    Am = L*Qx
+    #Somehow the operator is singular
+    @test_broken analytic_M = analytic_Am \ (reshape(M2, prod(s)) .- repeat(analytic_Ab, 3))
```
ChrisRackauckas, line 485: it's worth looking into this a bit and seeing if it really should be singular.

test/bc_coeff_compositions.jl
```diff
@@ -433,25 +445,63 @@ end
...
+    Dx=u₀*CenteredDifference{1}(1,4,Δx,N)
+    Dxx=Γ*CenteredDifference{1}(2,4,Δx,N)
+    Q=NeumannBC((2x[1]+2, 2x[end]+2), Δx, 6)
+
+    A = Dx*Q + Dxx*Q
+
+    y = A*(x.^2)
+    analytic_y = 2x.+2
+
+    @test_broken y ≈ analytic_y
```
ChrisRackauckas, line 506: hmm, does it get close?

ChrisRackauckas:
https://travis-ci.com/github/SciML/DiffEqOperators.jl/builds/186859179#L609 the failures look like they are just floating point tolerance issues.

src/derivative_operators/ghost_derivative_operator.jl
grahamas: I've confused myself.  The code I wrote just shaved extraneous padding off padded `u`s for calculation of the derivative in a single direction, and made sure `x` is compatible with some plausible `u` (with the caveat that right now it assumes 1 padding, as the code it replaces did).  I made #280, though.

grahamas:
Weird, I've run it several times locally and it always passes. What should I change?

ChrisRackauckas:
make it print out the maximum error: see if it's a floating point issue or if it's something bigger.

src/DiffEqOperators.jl
grahamas: no longer exported; renamed to `cartesian_to_linear`

test/bc_coeff_compositions.jl
grahamas:
```
julia> maximum(abs.(y .- analytic_y))
4.665789514226372
```

ChrisRackauckas: oof okay that's an issue, but it can be skipped for now.

grahamas:
Seems tiny (< 1e-15).  I'll replace the equalities by approxes, and a quick scan to see if this would be a problem anywhere else.

ChrisRackauckas:
I added a commit that adds a bit more safety too

ChrisRackauckas:
The PR passes for me locally, so I think this memory issue might do it.

src/derivative_operators/ghost_derivative_operator.jl
grahamas: The first two seem to be captured fine by defaults, or at least `3D_laplacian.jl` works out fine (granted it's 3D, so maybe I'm missing why the Matrix part of this dispatch is significant).

```
julia> (typeof((A*Q).ops[1]), typeof(u0))
(GhostDerivativeOperator{Float64,DerivativeOperator{Float64,1,false,Float64,StaticArrays.SArray{Tuple{5},Float64,1,5},StaticArrays.SArray{Tuple{1},StaticArrays.SArray{Tuple{6},Float64,1,6},1,1},Nothing,Nothing},ComposedMultiDimBC{Float64,RobinBC{Float64,V} where V<:AbstractArray{Float64,1},3,2}}, Array{Float64,3})

julia> which(LinearAlgebra.mul!, (typeof(u0), typeof((A*Q).ops[1]), typeof(u0)))
mul!(Y::AbstractArray, L::DiffEqBase.AbstractDiffEqLinearOperator, B::AbstractArray) in DiffEqBase at /home/graham/.julia/packages/DiffEqBase/zXRWE/src/operators/common_defaults.jl:17

julia> which(Base.:(*), (typeof((A*Q).ops[1]), typeof(u0)))
*(A::GhostDerivativeOperator{T1,E,F} where F<:DiffEqOperators.AbstractBC{T1} where E<:DiffEqBase.AbstractDiffEqLinearOperator{T1}, u::AbstractArray{T2,N} where N) where {T1, T2} in DiffEqOperators at /home/graham/.julia/dev/DiffEqOperators/src/derivative_operators/ghost_derivative_operator.jl:14
```
where the variables are as in `3D_laplacian.jl`

Caveat:  The output plot from `3D_laplacian.jl` looks pretty, but I don't actually know what it should look like. It's basically concentric 3D waves emanating from the center of space and then bouncing back from the edges of the cubic space and interfering with each other a bit.

ChrisRackauckas:
🎉 🎉 🎉

ChrisRackauckas:
You cannot even know how happy I am this is completed. This is a big step forward. I think we can say that the package now covers all it set out to do. Bug to fix? Sure. But the scope is there.

xtalax:
🎉 🎉 🎉 Thanks so much to all who worked to get this merged, I'll freely admit that I was completely stumped by some of the failing test cases at the point this was at by my last commit. Excited to see this being used, and  where the package goes from here

src/derivative_operators/ghost_derivative_operator.jl
xtalax: I can confirm that that is what it's supposed to look like, it's the wave eqn

ChrisRackauckas:
Let me know if either of you want to apply for a NumFOCUS small development grant to help keep pushing this forward.

src/derivative_operators/bc_operators.jl
xtalax: The stencil for PeriodicBC? It has the effect of putting the last value in u in the lower padding spot and the first value in the higher padding spot, as you'd expect in the periodic case

src/composite_operators.jl
xtalax: Yes, there were ambiguity issues.

PR #288, opened 2020-10-29
Upgrade to MTK 4.0

YingboMa:


ChrisRackauckas:
@jlperla would you know why these errors from SDO show up? Sounds like that package had a breaking change? We will merge the MTK update without handling that, but we need to get that test fixed.

PR #290, opened 2020-10-31
Remove need for calling SDO in tests

chenwilliam77:
This PR should address the breaking SDO tests. As @jlperla recommended, I copied and pasted the test output. I ran the tests successfully in Julia 1.5 on my personal laptop, too.


ChrisRackauckas:
Thanks!

PR #306, opened 2020-12-10
Issue 300 robin neumann bcs

### LINKED ISSUE #300: Add Neumann and Robin boundary conditions to MOLFiniteDifference
Currently, the MTK format only accepts boundary conditions of the form
```julia
bcs = [u(0,x) ~ -x*(x-1)*sin(x),
        u(t,0) ~ 0.0,
        u(t,1) ~ 0.0]
```
It would be good to add Neumann and Robin conditions, e.g.:
```julia
bcs = [u(0,x) ~ -x*(x-1)*sin(x),
        Dx(u(t,0)) ~ 0.0,
        u(t,1) + 2Dx(u(t,1)) ~ 0.0]
```
Comments on the API are welcome @ChrisRackauckas @emmanuellujan 

Also, it might be better to tackle #283 first, since this targets the same code

EDIT: I see NeuralPDE.jl does `Dx(u(t,0))` instead of `Dx(u)(t,0)`, so perhaps better to stick with that for consistency

tinosulzer:
Fixes #300


src/MOL_discretization.jl
```diff
@@ -6,29 +6,113 @@ struct MOLFiniteDifference{T} <: DiffEqBase.AbstractDiscretization
...
+        # Extract the variable from the lhs
+        if operation(lhs) isa Sym
+            # Dirichlet boundary condition
+            var = operation(lhs).name
+            α = 1.0
+            β = 0.0
+            bc_args = lhs.args
+        elseif operation(lhs) isa Differential
+            # Neumann boundary condition
+            var = operation(lhs.args[1]).name
```
ChrisRackauckas, line 37: Order of the derivative needs to be checked here?

src/MOL_discretization.jl
```diff
@@ -6,29 +6,113 @@ struct MOLFiniteDifference{T} <: DiffEqBase.AbstractDiscretization
...
-                j = 1
-            elseif isequal(bcs[i].lhs.args[2],domain.lower) # u(t,x=x_init)
-                j = 2
-            elseif isequal(bcs[i].lhs.args[2],domain.upper) # u(t,x=x_final)
-                j = 3
+        lhs = bcs[i].lhs
+        # Extract the variable from the lhs
+        if operation(lhs) isa Sym
+            # Dirichlet boundary condition
+            var = operation(lhs).name
```
ChrisRackauckas, line 31: I think in updated interface we want to be using `name(...)`. @YingboMa ?

src/MOL_discretization.jl
tinosulzer: ok, what's the command to get the order?

ChrisRackauckas: I think you want to check if `operation(operation(lhs)) isa Differential` and throw an error directly. That should be good for now since it has to be first order for Robin.

src/MOL_discretization.jl
tinosulzer: how do I get `name`? It's not in MTK

ChrisRackauckas: Sorry, `nameof`. It's from Base.

tinosulzer:
By the way, is there a way to have "Plots" installed without it being added to the Project.toml?

ChrisRackauckas:
You could add it as a test dependency, but plotting shouldn't be in the tests at all?

tinosulzer:
When I want to plot something to check it, I need to add `Plots`, but then it gets added to `Project.toml` so I have to remove it if I want to do `git add .`. I guess I could just add individual folders and files

ChrisRackauckas:
I always just `]dev` it and you never have that issue.

ChrisRackauckas:
Looks great. I think we'll want to split the docs apart to multiple example files and such soon, but that's a separate issue and not a reason to hold this up.

Issue #276, opened 2020-09-16
Support for Nonlinear Diffusion Problems


dangirsh:
Nonlinear diffusion problems have operators like:

![image](https://user-images.githubusercontent.com/1173905/92013400-909ac100-ed02-11ea-8528-64f877812c7b.png)

There doesn't seem to be existing support for constructing nonlinear operators like this.

A naive approach with the existing operators is

```julia
(∇ * bc2) (D_eff .* (∇ * bc1 * c)) 
```

, but it's not always possible to define `bc2`.

PR #327, opened 2021-01-31
Adding NonLinearDiffusion operator to derivate_operator.jl returning discretized function

mjsheikh:
This addition tries addressing #276.

It will return the required discretization as output given k (the diffusion coefficent) and u (the desired function) along with inner and outer differential orders.


ChrisRackauckas:
The test issue should be fixed https://github.com/SciML/SciMLBase.jl/pull/15.

This needs tests.

ChrisRackauckas (approved):

examples/Fast_Diffusion.jl
```diff
@@ -0,0 +1,46 @@
...
+u0 = u_analytic.(knots,0.0)
+
+t0 = 0.0
+t1 = 1.0
+
+function f(u,p,t)
+    bc = DirichletBC(exp(-t),(1.0 + exp(2*t))^(-0.5))
+    l = bc*u
+    k = l.^(-2)                        # Diffusion Coefficient
+    NonLinearDiffusion(n,m,approx_ord,k,l,h,nknots)
```
ChrisRackauckas, line 37: Can it be represented as an operator?

src/derivative_operators/derivative_operator.jl
```diff
@@ -26,6 +26,31 @@ struct DerivativeOperator{T<:Real,N,Wind,T2,S1,S2<:SArray,T3,F} <: AbstractDeriv
...
+    else 
+        discretization += q[2:(nknots + 1)].*(CenteredDifference(second_differential_order,approx_order,dx,nknots)*p); 
+    end
+    
+    for l = 1:(second_differential_order - 1)
+        discretization += binomial(second_differential_order,l)*(CenteredDifference(l + first_differential_order,approx_order,dx,nknots)*q).*(CenteredDifference(second_differential_order - l,approx_order,dx,nknots)*p);
+    end
+    
+    discretization += (CenteredDifference(first_differential_order + second_differential_order,approx_order,dx,nknots)*q).*p[2:(nknots + 1)];
+    return discretization
```
ChrisRackauckas, line 50: There only seems to be an out of place operation? It would be good to have a `mul!` definition and then an allocating version that uses the non-allocating one. Refer to https://www.youtube.com/watch?v=M2i7sSRcSIw or https://mitmath.github.io/18337/lecture2/optimizing on non-allocating operators.

examples/Fast_Diffusion.jl
ChrisRackauckas: For reference: 

http://ucidatascienceinitiative.github.io/IntroToJulia/Html/IntermediateProblems#Operator-Problem
http://ucidatascienceinitiative.github.io/IntroToJulia/Html/IntermediateProblemAnswers#Operator-Problem

ChrisRackauckas:
https://diffeq.sciml.ai/stable/features/diffeq_operator/ defines our operator format which is an extension of the simple examples posted above. I think this is all mathematically correct but needs a little bit of work to match the standard operator API.

src/derivative_operators/derivative_operator.jl
```diff
@@ -26,6 +26,29 @@ struct DerivativeOperator{T<:Real,N,Wind,T2,S1,S2<:SArray,T3,F} <: AbstractDeriv
     coeff_func              :: F
 end
 
+struct NonLinearDiffusion! end
```
ChrisRackauckas, line 29: if it's a function then it shouldn't be capitalized like a struct, or need the struct at all.

examples/Fast_Diffusion.jl
```diff
@@ -0,0 +1,46 @@
+# # Fast-Diffusion Problem
```
ChrisRackauckas, line 1: this needs to be added to the tests. We really need to remove the examples folder and instead just write tutorials in the docs too.

test/Fast_Diffusion.jl
```diff
@@ -0,0 +1,50 @@
...
+# Here we consider a fast diffusion problem with Dirichlet BCs on unit interval:
+# ∂ₜu = ∂ₓ(k*∂ₓu) 
+# k = 1/u²   
+# u(x=0,t) = exp(-t)
+# u(x=1,t) = 1/(1.0 + exp(2t))
+# u(x, t=0) = u₀(x)
+using Test
+using DiffEqOperators, OrdinaryDiffEq
+
+@testset "1D Nonlinear fast diffusion equation" begin
```
ChrisRackauckas, line 13: it needs to be called from runtests.jl

ChrisRackauckas:
Just the caps issue and the lack of an out of place dispatch to fix.

ChrisRackauckas:
Great thanks!

ChrisRackauckas:
Could you follow up with adding this to the documentation?

mjsheikh:
> Could you follow up with adding this to the documentation?

Modifications should be done in **derivate_operators.md** or a new file needs to be created?

ChrisRackauckas:
I think a nonlinear derivative operators page would be good

Issue #331, opened 2021-02-09
Sparse matrix of laplacian with periodic BC


rveltz:
Hi,

I am surprised this does not work. I am doing it wrong?

```julia
using SparseArrays
using DiffEqOperators

function DiffOp(N, lx)
	hx = 2lx/N
	# D   = CenteredDifference(2, 2, hx, N)
	D2x = CenteredDifference(2, 2, hx, N)
	Qx = PeriodicBC(typeof(hx))
	Δ = sparse(D2x * Qx)
	return Δ
end
DiffOp(100, 1.)
```

returns

```julia
julia> DiffOp(100, 1.)
ERROR: DimensionMismatch("")
Stacktrace:
 [1] mul!(C::Vector{Float64}, A::SparseMatrixCSC{Float64, Int64}, B::Vector{Float64}, α::Bool, β::Bool)
   @ SparseArrays /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/linalg.jl:29
 [2] *
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/linalg.jl:48 [inlined]
 [3] SparseMatrixCSC(A::GhostDerivativeOperator{Float64, DerivativeOperator{Float64, 1, false, Float64, StaticArrays.SVector{3, Float64}, StaticArrays.SVector{0, StaticArrays.SVector{4, Float64}}, Nothing, Nothing}, PeriodicBC{Float64}}, N::Int64)
   @ DiffEqOperators ~/.julia/packages/DiffEqOperators/otkmY/src/derivative_operators/concretization.jl:729
 [4] sparse
   @ ~/.julia/packages/DiffEqOperators/otkmY/src/derivative_operators/concretization.jl:739 [inlined]
 [5] DiffOp(N::Int64, lx::Float64)
   @ Main ./REPL[2]:6
 [6] top-level scope
   @ REPL[5]:1
```

PR #334, opened 2021-02-15
dimensional corrections for PeriodicBC

mjsheikh:
Fixes for #331


ChrisRackauckas:
Can this get a test in https://github.com/SciML/DiffEqOperators.jl/blob/master/test/ghost_derivative_operators_interface.jl ?

PR #340, opened 2021-03-02
tests added

mjsheikh:


test/runtests.jl
```diff
@@ -24,7 +24,7 @@ if GROUP == "All" || GROUP == "Interface"
     @time @safetestset "2nd order check" begin include("2nd_order_check.jl") end
     @time @safetestset "Non-linear Diffusion" begin include("Fast_Diffusion.jl") end
     #@time @safetestset "KdV" begin include("KdV.jl") end # KdV times out and all fails
```
ChrisRackauckas, line 26: is there a reason why KdV wasn't added back?

mjsheikh: Because tests would be broken for the 2-Soliton case

ChrisRackauckas: just mark that section?

mjsheikh: If you say so :D

Issue #97, opened 2019-06-03
Implement Biased Upwind Operators


ajozefiak:
Currently the UpwindOperator type implements a pure upwind scheme (in terms of it its stencils). Additionally, biased upwind schemes which utilize a specific number of off-sided points would be a useful functionality. For instance, the following paper describes a biased upwind scheme utilizing a single off-sided point: https://www.sciencedirect.com/science/article/pii/S0378475401002889 

The default number of off-sided points should be zero, defaulting back to the pure upwind scheme that is currently implemented.

PR #347, opened 2021-03-19
Adding Biased Upwind Operator

mjsheikh:
Implementation of a Biased Upwind Operator ( #97 )
TODO : Doc changes once the code gets approved.

### DIFF
--- a/src/derivative_operators/derivative_operator.jl
+++ b/src/derivative_operators/derivative_operator.jl
@@ -11,7 +11,7 @@
 
 index(i::Int, N::Int) = i + div(N, 2) + 1
 
-struct DerivativeOperator{T<:Real,N,Wind,T2,S1,S2<:SArray,T3,F} <: AbstractDerivativeOperator{T}
+struct DerivativeOperator{T<:Real,N,Wind,T2,S1,S2,S3<:SArray,T3,F} <: AbstractDerivativeOperator{T}
     derivative_order        :: Int
     approximation_order     :: Int
     dx                      :: T2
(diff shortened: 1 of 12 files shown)


src/derivative_operators/derivative_operator.jl
```diff
@@ -182,69 +187,85 @@ julia> Array(L2 * Q)[1]
...
-        high_boundary_coefs,coefficients,coeff_func
+        high_boundary_coefs,offside,coefficients,coeff_func
         )
 end
 
 # TODO implement the non-uniform grid
 function UpwindDifference{N}(derivative_order::Int,
                           approximation_order::Int, dx::AbstractVector{T},
-                          len::Int, coeff_func=nothing) where {T<:Real,N}
+                          len::Int, offside::Int, coeff_func=nothing) where {T<:Real,N}
```
ChrisRackauckas, line 233: should be a keyword argument

mjsheikh:
Will add tests shortly

test/heat_eqn.jl
```diff
@@ -33,7 +33,7 @@ end
     dx = 2π/(N-1)
     x = collect(-pi : dx : pi)
     u0 = @. -(x - 0.5)^2 + 1/12
-    B = CenteredDifference(1,2,dx,N)
+    B = CenteredDifference(1,2,dx,N-2)
```
ChrisRackauckas, line 36: why the size change?

mjsheikh: > why the size change?

Its operating on a Vector with 512 grid points(including boundaries).
That was working somehow on Centered difference but Upwind was returning dimension mismatch error.

PR #361, opened 2021-03-28
Issue 351 neumann bcs

tinosulzer:
Fixes sign error in the boundary condition replacement for the left-hand side

Also hopefully makes MOL_discretization.jl more readable to non-cyborgs

Fixes https://github.com/SciML/DiffEqOperators.jl/issues/351


src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -45,19 +47,29 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+                    for (j, depvar) in enumerate(depvars)]
+        # Create list of all the symbolic Differential terms evaluated at boundary e.g. Differential(x)(u(t,0))
+        subderivar(depvar,s) = substitute.((Differential(s)(depvar),),edgevals)
+        # Create map of symbolic Differential terms with symbolic spatially discretized terms
+        depvarderivmaps = reduce(vcat,[subderivar(depvar, s) .=> derivars[i]
+                                       for (i, depvar) in enumerate(pdesys.depvars) for s in nottime])
    else
-       # TODO: Fix Neumann and Robin on higher dimension
-       depvarderivmaps = []
+        # Higher order system
```
ChrisRackauckas, line 70: dimension

test/MOL/MOL_1D_Linear_Diffusion.jl
```diff
@@ -265,159 +263,15 @@ end
...
-    # Convert the PDE problem into an ODE problem
-    prob = discretize(pdesys,discretization)
-
-    # Solve ODE problem
-    sol = solve(prob,Tsit5(),saveat=0.1)
-    x = prob.space[2]
-    t = sol.t
-
-    x = dx[2:end-1]
+    x = (-1:dx:1)[2:end-1]
```
ChrisRackauckas, line 266: With the deletions we should uncomment this test now?

tinosulzer: Yes this test file now works. Except test 6 which creates a DAE system (I guess it could work with a DAE solver, didn't try), and test 2 with the nonlinear laplacian

ChrisRackauckas: I'll take a quick look.

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -45,19 +47,29 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+        left_idxs = central_neighbor_idxs(CartesianIndex(2),1)[1:2]
+        right_idxs(j) = central_neighbor_idxs(CartesianIndex(length(space[j])-1),1)[end-1:end]
+        # Constructs symbolic spatially discretized terms of the form e.g. au₂ - bu₁
+        derivars = [[dot(left_weights(j),depvar[left_idxs]), dot(right_weights(j),depvar[right_idxs(j)])]
+                    for (j, depvar) in enumerate(depvars)]
+        # Create list of all the symbolic Differential terms evaluated at boundary e.g. Differential(x)(u(t,0))
+        subderivar(depvar,s) = substitute.((Differential(s)(depvar),),edgevals)
+        # Create map of symbolic Differential terms with symbolic spatially discretized terms
+        depvarderivmaps = reduce(vcat,[subderivar(depvar, s) .=> derivars[i]
+                                       for (i, depvar) in enumerate(pdesys.depvars) for s in nottime])
```
tinosulzer, line 68: this whole loop is only for the 1D case but is written more generally - should it be simplified?

ChrisRackauckas:
Note that `@test_broken` is highly preferred to `@test_skip`. I made the change there.

I fixed Test 6. That's just one that is a bit nasty right now because we don't have https://github.com/SciML/DiffEqOperators.jl/issues/357, so I did `symbolic_discretize` to see that the order you get after `structural_simplify` is a bit odd. Of course, the moment `sol[u]` is working then it'll all be good, and really the issue here is that when using a symbolic system you shouldn't ever assume representation. So for now I hardcoded what it simplifies to and we should update that in the future.

ChrisRackauckas (approved):

src/MOLFiniteDifference/MOL_discretization.jl
ChrisRackauckas: I think simplifying it is the wrong direction: we should just finish making it generic to dimensionality. https://github.com/SciML/DiffEqOperators.jl/issues/350 has some starter code in that direction. I just thought it was unproductive to hold up getting some other contributors going to try and finish that, knowing that I have more than a few grant proposals to write.

Issue #352, opened 2021-03-28
Allow order choice in central difference handling of MOLFiniteDifference


ChrisRackauckas:
After https://github.com/SciML/DiffEqOperators.jl/pull/349 it's hardcoded. One needs to generalize the lines:

```julia
        central_neighbor_idxs(i,j) = [i+CartesianIndex([ifelse(l==j,-1,0) for l in 1:length(nottime)]...),i,i+CartesianIndex([ifelse(l==j,1,0) for l in 1:length(nottime)]...)]
        central_weights(i,j) = DiffEqOperators.calculate_weights(2, 0.0, [space[j][i[j]-1],space[j][i[j]],space[j][i[j]+1]])
        central_deriv_rules = [(Differential(nottime[j])^2)(pdesys.depvars[k]) => dot(central_weights(i,j),depvars[k][central_neighbor_idxs(i,j)]) for j in 1:length(nottime), k in 1:length(pdesys.depvars)]
```

Note from the Fornberg that the order is determined by the number of stencil points:

https://github.com/SciML/DiffEqOperators.jl/blob/v4.22.0/src/derivative_operators/fornberg.jl#L2-L16

In the `CenteredDifference` operator you can see how it's done:

https://github.com/SciML/DiffEqOperators.jl/blob/v4.22.0/src/derivative_operators/derivative_operator.jl#L128

tinosulzer:
What does `x0` do in `calculate_weights`? I see you just set it to 0 everywhere

ChrisRackauckas:
It's mostly for type dispatch at this point. That needs to get refactored a bit, since it should probably be pulling that from the stencil types.

tinosulzer:
This can be closed

PR #365, opened 2021-03-30
#352 allow higher-order central difference; add MOL_2D test with Dirichlet

tinosulzer:
Uses CartesianIndices to apply higher-order central difference method.
Needs special handling for cases close to edges (for example index(2) - 2 gives Index(0) which does exist, so instead we need to use indices 1-5 to get 4th order). Fornberg does this for you but I'm not 100% sure it comes out as being the right order near the boundaries, since forward differences need more points to get the same accuracy https://en.wikipedia.org/wiki/Finite_difference_coefficient


ChrisRackauckas (approved):

ChrisRackauckas:
There's some theory on this IIRC. With ghost point boundaries at order 1 you can prove you still get a consistent order 2 discretization because of rate at which the boundaries fall off because of their lower dimensionality. Because of how the proof works, I'd be pretty sure it extends to higher order.

ChrisRackauckas:
You could take non-centered stencils near the boundary, which is what the derivative operators do, but that seems like too much work with little gain.

tinosulzer:
This actually does the non-centered stencils near the boundary (turns out this is what the `x0` in fornberg is for, too) - I was just wondering whether we'd need an extra point at the boundary to get the same accuracy but agree the gain is minimal https://github.com/SciML/DiffEqOperators.jl/blob/45f79f2d8036d457356b3cc7cf43ea017c955228/src/MOLFiniteDifference/MOL_discretization.jl#L118-L122

Issue #354, opened 2021-03-28
Nonlinear Laplacian Handling in MOLFiniteDifference


ChrisRackauckas:
This requires a rule application so that the substitution can look for an arbitrary `D(a * D(u))` and rewrite with `a` . Some of the tricks for this are shown in the upwinding section of the code, but this requires an additional feature in the SymbolicUtils rules for interpolation into rules.

@shashi @YingboMa

tinosulzer:
Can you elaborate more on this? How would you rewrite something like
```julia
D(4 * exp(- 3 * u) * D(u))
```
?

ChrisRackauckas:
The rewrite rule is essentially a regex, so it would catch `D(4 * exp(- 3 * u) * D(u))` as `D(a * D(u))` with `a = 4 * exp(- 3 * u)`. The rewrite would then be to use the same rules as @mjsheikh 's PR https://github.com/SciML/DiffEqOperators.jl/pull/327, which is the generalized divergence form (scheme 2 of https://math.stackexchange.com/questions/1949795/explicit-finite-difference-scheme-for-nonlinear-diffusion/1950816 ) which is a bit more stable than reducing to two terms. So that would then be:

```julia
(a(x+1/2) * (u[i+1] - u[i]) - a(x-1/2) * (u[i] - u[i-1]) / dx^2
```

which you can then generalize as in that PR with Fornberg. Now if it's just `a(x,t,u) == a(x,t)` then you're done. If it's dependent on `u` then you have another complication. But the proper derivation of finite difference schemes comes from differentiation of interpolating polynomials:

https://mitmath.github.io/18337/lecture14/pdes_and_convolutions

So implicitly to the scheme there us a polynomial that is fit and would thus give the evaluation of `u(x+1/2)`. For example, at second order it is:

![Capture](https://user-images.githubusercontent.com/1814174/113418402-6a6dfa00-9393-11eb-8008-5e777c6fe4cc.PNG)

Evaluating that at u(x+1/2) thus gives an approximation to `u` that retains the 2nd order accuracy in the equation and uses precisely the same points.

[...]

PR #371, opened 2021-04-13
Nonlinear Laplacian Handling in MOLFiniteDifference (issue #354)

emmanuellujan:
Fixes part of #354


ChrisRackauckas:
Is there a reason why this is commented?

emmanuellujan:
I am working with some test files in my computer that isolate the portion of code that performs this particular parsing. This code is not integrated to MOL_discretization.jl yet, first I need to solve some issues with the automatic generation of rules involving Differentials. Because I wanted to upload some progress, I added the portion of code of interest at about the place where it should be integrated, but as a comment.

emmanuellujan:
I added code to handle cases like the following: 
- Dt(u(t,x)) ~ Dx(u(t,x)^(-1) * Dx(u(t,x)))
- Dt(u(t,x)) ~ Dx(u(t,x)^2 * Dx(u(t,x)))
- Dt(u(t,x)) ~ Dx(1. / (1. + u(t,x)^2) * Dx(u(t,x)))
- Dt(u(t,x)) ~ Dx(1. / (u(t,x)^2 - 1.) * Dx(u(t,x)))
- Dt(u(t,x,y)) ~ Dx( (u(t,x,y)^2 / exp(x+y)^2 + sin(x+y+4t)^2)^0.5 * Dx(u(t,x,y))) +
                       Dy( (u(t,x,y)^2 / exp(x+y)^2 + sin(x+y+4t)^2)^0.5 * Dy(u(t,x,y)))

To do this I partially used the SymbolicUtils `@rules`.

Also, I added new test cases for all the examples above.

The function `g`  (see issue "https://github.com/SciML/DiffEqOperators.jl/issues/354") needs to be improved. Currently a basic approximation is implemented. I would like to solve this in another PR. Now I would like to address convection and convection-diffusion.

emmanuellujan:
@ChrisRackauckas if you agree, you can merge this PR.

ChrisRackauckas:
It's WIP and has merge conflicts?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -1,4 +1,7 @@
 using ModelingToolkit: operation, istree, arguments
+#using SymbolicUtils
```
ChrisRackauckas, line 2: ?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -24,6 +27,9 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
         dx = discretization.dxs[findfirst(dxs->isequal(x.val, dxs[1].val),discretization.dxs)][2]
         dx isa Number ? (xdomain.domain.lower:dx:xdomain.domain.upper) : dx
     end
+    dxs = map(nottime) do x
```
ChrisRackauckas, line 30: rebase?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -126,18 +135,47 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+        r_mid_dep(i, j, k, l) = [pdesys.depvars[k] => g(i, j, k, l) for k in 1:length(pdesys.depvars)]
+        # Independent variable rules
+        r_mid_indep(i, j, l) = [nottime[j] => iv_mid(i, j, l) for j in 1:length(nottime)]
+        # Replacement rules: new approach
+        rules = [@rule ($(Differential(iv))(*(~~a, $(Differential(iv))(dv), ~~b))) =>
+                 dot([Num(substitute(substitute(*(~~a..., ~~b...), r_mid_dep(II, j, k, -1)), r_mid_indep(II, j, -1))),
+                      Num(substitute(substitute(*(~~a..., ~~b...), r_mid_dep(II, j, k, 1)), r_mid_indep(II, j, 1)))],
+                     [-b1(II, j, k), b2(II, j, k)])
+                 for (j, iv) in enumerate(nottime) for (k, dv) in enumerate(pdesys.depvars)]
+        rhs_arg = (SymbolicUtils.operation(eq.rhs) == +) ? SymbolicUtils.arguments(eq.rhs) : [eq.rhs]
```
ChrisRackauckas, line 164: why do you need to explicitly specialize on `+`?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -126,18 +135,47 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+        lhs_arg = (SymbolicUtils.operation(eq.lhs) == +) ? SymbolicUtils.arguments(eq.lhs) : [eq.lhs]
+        nonlinlap_rules = []
+        for t in vcat(lhs_arg,rhs_arg)
+            for r in rules
+                if r(t) != nothing
+                    push!(nonlinlap_rules, t => r(t))
+                end
+            end
+        end
```
ChrisRackauckas, line 173: why does this need to be done manually? @YingboMa @shashi don't the rules check the variations?

emmanuellujan:
> Is there a reason why this is commented?

I update my fork from time to time, I estimate that the conflicts were generated because of that. Unfortunately I do not see an option in github that allows me to resolve them. If there is such an option, and you could point me to it, I would be happy to resolve them.

emmanuellujan:
> It's WIP and has merge conflicts?

I update my fork from time to time, I estimate that the conflicts were generated because of that. Unfortunately I do not see an option in github that allows me to resolve them. If there is such an option, and you could point me to it, I would be happy to resolve them.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Thanks for pointing this out to me, I just removed those comments.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: I am not sure about the meaning of the question.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: My goal is to run classical convection-diffusion equations. To do so, in this PR I addressed summations involving nonlinear Laplacians, needed in diffusive terms.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Unfortunately, I did not find a way to fully use SymbolicUtils to solve this issue. So I solved a part of it manually. An explanation of the my issue can be found in the following document.: https://docs.google.com/document/d/1qbWokDYXuO5GV3ilI8R4fWWnvkBkrpQd3H6d7l9DrC4/edit

tinosulzer:
I think you need write access to resolve conflicts on github, but if you merge master into your local branch you will see them and be able to resolve them

emmanuellujan:
> I think you need write access to resolve conflicts on github, but if you merge master into your local branch you will see them and be able to resolve them

Thanks for the advice. In summary this is what I did:
```
git clone https://github.com/emmanuellujan/DiffEqOperators.jl/
cd DiffEqOperators.jl
git remote add upstream https://github.com/SciML/DiffEqOperators.jl/
git fetch upstream
git checkout master
git merge upstream/master
git checkout fix-issue-354
git merge master
# Solve conflict in editor 
git add src/MOLFiniteDifference/MOL_discretization.jl
git push
```

emmanuellujan:
@ChrisRackauckas @tinosulzer

src/MOLFiniteDifference/MOL_discretization.jl
tinosulzer: edit the PR, change the base branch from SciML:master to anything else, then edit again and change back to SciML:master. This will get rid of changes like this which are just git struggling to keep track (maybe there is a smarter way haha)

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -118,12 +122,15 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
         # TODO: Generalize central difference handling to allow higher even order derivatives
         # The central neighbour indices should add the stencil to II, unless II is too close
         # to an edge in which case we need to shift away from the edge
+
         # Calculate buffers
-        I1 = oneunit(first(indices))
-        Imin = first(indices) + I1 * (order÷2)
-        Imax = last(indices) - I1 * (order÷2)
+        # TODO: check central_neighbor_idxs. It does not work like the old implementation.
```
tinosulzer, line 127: hmm, weird. I don't mind taking a look at this again after it's merged.

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -132,18 +139,48 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+        for t in vcat(lhs_arg,rhs_arg)
+            for r in rules
+                if r(t) != nothing
+                    push!(nonlinlap_rules, t => r(t))
+                end
+            end
+        end
+
+        rules = vcat(vec(nonlinlap_rules),
+                     vcat(vec(central_deriv_rules),valrules))
```
tinosulzer, line 180: why does this need 2 separates vcats?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -132,18 +139,48 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+        #                         for j in 1:nspace, k in 1:length(pdesys.depvars)]
+
+        ## Discretization of non-linear laplacian. 
+        # d/dx( a du/dx ) ~ (a(x+1/2) * (u[i+1] - u[i]) - a(x-1/2) * (u[i] - u[i-1]) / dx^2
+        b1(i, j, k) = dot(reverse_weights(i, j), depvarsdisc[k][central_neighbor_idxs(i, j)[1:2]]) / dxs[j]
+        b2(i, j, k) = dot(forward_weights(i, j), depvarsdisc[k][central_neighbor_idxs(i, j)[2:3]]) / dxs[j]
+        # TODO: improve interpolation of g(x) = u(x) for calculating u(x+-dx/2)
+        g(i, j, k, l) = sum([depvarsdisc[k][central_neighbor_idxs(i, j)][s] for s in (l == 1 ? [2,3] : [1,2])]) / 2.
+        # iv_mid returns middle space values. E.g. x(i-1/2) or y(i+1/2).
+        iv_mid(i, j, l) = (space[j][i[j]] + space[j][i[j]+l]) / 2.0
```
tinosulzer, line 157: note harmonic mean might be more accurate http://web.cecs.pdx.edu/~gerry/class/ME448/lecture/pdf/CVFDdiffusion2D.pdf

tinosulzer (approved):
I am ok merging this as something that works for now so we can move forward. But I am not sure this is the most accurate approach: currently we do D(u_i+1/2, x_i+1/2) (i.e. function of the average) but usually I have seen D_i+1/2 (i.e. average of the function). These can be quite different if D is nonlinear. Would need to dig more into the numerical analysis to be sure which is best

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Done :)

emmanuellujan:
> I am ok merging this as something that works for now so we can move forward. But I am not sure this is the most accurate approach: currently we do D(u_i+1/2, x_i+1/2) (i.e. function of the average) but usually I have seen D_i+1/2 (i.e. average of the function). These can be quite different if D is nonlinear. Would need to dig more into the numerical analysis to be sure which is best

Currently a basic approximation is implemented. I would like to solve this in another PR.  In the first page of this [draft](https://docs.google.com/document/d/1qbWokDYXuO5GV3ilI8R4fWWnvkBkrpQd3H6d7l9DrC4/edit) you can find the definition of g based on a degree 2 polynomial in a 3D case. If this is the correct definition, I would need to know which stencil nodes should I use. The "cross" gives me 7, but I need 10 to fit the coefficients, I think that the crossed derivatives could be avoided to use just 7 nodes.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Thanks!

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Thanks for the note. I added to my [draft](https://docs.google.com/document/d/1qbWokDYXuO5GV3ilI8R4fWWnvkBkrpQd3H6d7l9DrC4/edit).

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: Thanks, I am using one vcat now :)

ChrisRackauckas:
>Currently a basic approximation is implemented. I would like to solve this in another PR. In the first page of this draft you can find the definition of g based on a degree 2 polynomial in a 3D case. If this is the correct definition, I would need to know which stencil nodes should I use. The "cross" gives me 7, but I need 10 to fit the coefficients, I think that the crossed derivatives could be avoided to use just 7 nodes.

You split it to the 3 terms Del u = u_xx + u_yy + u_zz  and then you have 2 unique coefficients per, + 1 for the middle. Otherwise you have to have the corner terms, which you could discretize to include the corners.

src/MOLFiniteDifference/MOL_discretization.jl
ChrisRackauckas: @YingboMa can you double check this?

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -118,12 +122,15 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
         # to an edge in which case we need to shift away from the edge
+
         # Calculate buffers
-        I1 = oneunit(first(indices))
-        Imin = first(indices) + I1 * (order÷2)
-        Imax = last(indices) - I1 * (order÷2)
+        # TODO: check central_neighbor_idxs. It does not work like the old implementation.
+        #I1 = oneunit(first(indices))
+        #Imin = first(indices) + I1 * (order÷2)
+        #Imax = last(indices) - I1 * (order÷2)
```
ChrisRackauckas, line 130: just remove

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -118,12 +122,15 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
         # Use max and min to apply buffers
-        central_neighbor_idxs(II,j) = stencil(j) .+ max(Imin,min(II,Imax))
+        # central_neighbor_idxs(II,j) = stencil(j) .+ max(Imin,min(II,Imax))
```
ChrisRackauckas, line 132: ? just remove if there's a new definition

tinosulzer: it's back to the old definition which ignores the discretization order (I think) i.e. breaks #365 but not enough to cause tests to fail

ChrisRackauckas:
>I am ok merging this as something that works for now so we can move forward. But I am not sure this is the most accurate approach: currently we do D(u_i+1/2, x_i+1/2) (i.e. function of the average) but usually I have seen D_i+1/2 (i.e. average of the function). These can be quite different if D is nonlinear. Would need to dig more into the numerical analysis to be sure which is best

That would make the approximation O(dx) instead of O(dx^2), but that can be improved in a subsequent PR. Let's get an issue open for this.

src/MOLFiniteDifference/MOL_discretization.jl
ChrisRackauckas: weird.

ChrisRackauckas: Are you looking into this?

tinosulzer: I can take a look tomorrow

emmanuellujan:
> > Currently a basic approximation is implemented. I would like to solve this in another PR. In the first page of this draft you can find the definition of g based on a degree 2 polynomial in a 3D case. If this is the correct definition, I would need to know which stencil nodes should I use. The "cross" gives me 7, but I need 10 to fit the coefficients, I think that the crossed derivatives could be avoided to use just 7 nodes.
> 
> You split it to the 3 terms Del u = u_xx + u_yy + u_zz and then you have 2 unique coefficients per, + 1 for the middle. Otherwise you have to have the corner terms, which you could discretize to include the corners.

I know, but if you want to fit a degree 2 polynomial in 3D you need 10 coefficients, ergo 10 nodes. If you use a typical stencil (without the corners), you have 7 nodes, ergo you need to neglect some terms of the polynomial. I think the cross derivative terms could be removed to this end. Another option is to use the corners. In this case, you have 15 nodes available, but you only need 10, so you need to chose which corners to use. Probably any choice in which the nodes are sufficiently distributed is enough.

src/MOLFiniteDifference/MOL_discretization.jl
tinosulzer: The only thing that is broken with the `stencil` approach is "Test 00: Dt(u(t,x,y)) ~ Dxx(u(t,x,y)) + Dyy(u(t,x,y))" in the `MOL_2D_Diffusion.jl` test with orders 4 and 6. The test breaks because the solution is not accurate enough, not a syntax error. The reason it works with the old syntax is that it just ignores `discretization.centered_order` and pretends you're always asking for order 2. @emmanuellujan if you only test for order 2 in that test then you can use the syntax you commented out, and we should open an issue for the order 4 and 6 in 2D.

emmanuellujan:
> > I am ok merging this as something that works for now so we can move forward. But I am not sure this is the most accurate approach: currently we do D(u_i+1/2, x_i+1/2) (i.e. function of the average) but usually I have seen D_i+1/2 (i.e. average of the function). These can be quite different if D is nonlinear. Would need to dig more into the numerical analysis to be sure which is best
> 
> That would make the approximation O(dx) instead of O(dx^2), but that can be improved in a subsequent PR. Let's get an issue open for this.

Okay, we need to wait until this PR is merged to post the new issue.

Note: currently issue #371 contains this "sub-issue".

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: I understand that it is better not to remove those comments for now.

src/MOLFiniteDifference/MOL_discretization.jl
emmanuellujan: See below

src/MOLFiniteDifference/MOL_discretization.jl
tinosulzer: actually I'll just open a PR to your fork

emmanuellujan: Oh, we posted almost at the same time, I did not see your comment when I posted mine. I will check what you mention. Thank you :)

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -165,18 +169,48 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+                 dot([Num(substitute(substitute(*(~~a..., ~~b...), r_mid_dep(II, j, k, -1)), r_mid_indep(II, j, -1))),
+                      Num(substitute(substitute(*(~~a..., ~~b...), r_mid_dep(II, j, k, 1)), r_mid_indep(II, j, 1)))],
+                     [-b1(II, j, k), b2(II, j, k)])
+                 for (j, iv) in enumerate(nottime) for (k, dv) in enumerate(depvars)]
+        rhs_arg = (SymbolicUtils.operation(eq.rhs) == +) ? SymbolicUtils.arguments(eq.rhs) : [eq.rhs]
+        lhs_arg = (SymbolicUtils.operation(eq.lhs) == +) ? SymbolicUtils.arguments(eq.lhs) : [eq.lhs]
+        nonlinlap_rules = []
+        for t in vcat(lhs_arg,rhs_arg)
+            for r in rules
+                if r(t) !== nothing
```
ChrisRackauckas, line 203: ?

ChrisRackauckas: what exactly is this doing?

emmanuellujan: This code allows the non linear laplacian rules (`rules`) to be used as an argument of the function `substitute`, in an analogous way to what is done with the central derivative rules (see the code below). The first `for` iterates over the terms of the equation. The second `for` iterates over the non-linear laplacian rules (thus, a version of the rules that can not be used as an argument of the function `substitute`). If the current rule can be applied to the current term, a new rule in the correct format is generated.
Another option is not to use the `substitute` function to apply the nonlinear laplacian rules, but if we want to be consistent with what was done with the central derivative rules, In think that we have to make a transformation. Suggestions are welcome :)

ChrisRackauckas: No but what is `r(t) !== nothing` actually doing, and is that public API?

emmanuellujan: `r(t)` is applying the rule `r` to the term `t`, if the rule does not match with the term it returns `nothing`, otherwise it returns the transformed term. When we use `@rule`  (see code above) we need to work with the rule as it were a function, therefore we cannot directly use `substitute` as it is used with the central derivative rules. 

It is in the public API ("https://symbolicutils.juliasymbolics.org/api/"):

> `rule LHS => RHS`
> Creates a `Rule` object. A rule object is callable, and takes an expression and rewrites it if it matches the LHS pattern to the RHS pattern, returns nothing otherwise.

emmanuellujan: `r(t) !== nothing` is asking if the current rule can be applied to the current term.

ChrisRackauckas: It should probably just be `!=` but that's not a huge deal.

emmanuellujan: Great, I will fix it in the next PR :)

tinosulzer: that's my bad, VScode had a blue underline under `!=` which disappeared when I changed to `!==`

Issue #367, opened 2021-04-01
Finite differences for Laplacian in a spherical domain


tinosulzer:
https://web.mit.edu/braatzgroup/analysis_of_finite_difference_discretization_schemes_for_diffusion_in_spheres_with_variable_diffusivity.pdf

Includes nonlinear diffusion

PR #373, opened 2021-04-16
Issue 367 sphere

tinosulzer:
Add discretization of the Laplacian in spherical coordinates

TODO: Nonlinear Laplacian, leaving #367 open


src/DiffEqOperators.jl
```diff
@@ -62,6 +63,7 @@ export AbstractDerivativeOperator, DerivativeOperator,
        CenteredDifference, UpwindDifference, nonlinear_diffusion, nonlinear_diffusion!
 export DirichletBC, Dirichlet0BC, NeumannBC, Neumann0BC, RobinBC, GeneralBC, MultiDimBC, PeriodicBC,
        MultiDimDirectionalBC, ComposedMultiDimBC
+export AxisymmetricSphereDomain
```
ChrisRackauckas, line 66: can we make use of DomainSets.jl here? That's what I'm hoping to standardize around.

tinosulzer: Yeah this was only meant as a temporary hack until the DomainSets.jl interface is clearer - which seems like it would first require changes to [ModelingToolkit's `domain.jl`](https://github.com/SciML/ModelingToolkit.jl/blob/e1b8e40c37817643b34199431d56c4b70c7e77b4/src/domains.jl)? Since `VarDomainPairing` expects an `AbstractDomain`.

tinosulzer: Opened https://github.com/SciML/ModelingToolkit.jl/pull/976

tinosulzer:
I removed the spherical domain stuff and am just using `IntervalDomain` for now, looking for the form `x^2*Dx(1/x^2*Dx(u(t,x)))`

ChrisRackauckas:
The tests aren't using spherical domains and instead are just writing out the spherical laplacian in cartesian coordinates?

tinosulzer:
Yes

PR #375, opened 2021-04-19
Operators for Vector Calculus

mjsheikh:
This would ''just'' work for AbstractArray gradient computation.
But I guess it gives the idea for the way forward. Pursue the vector elements as corresponding entries in various dimensions.
Thoughts for the overall structure @ChrisRackauckas ?


ChrisRackauckas:
I think this is good? I would like to see curl implemented as a good test of whether this structure can work.

mjsheikh:
Right. Is this the best way to store a discretized vector?

ChrisRackauckas:
What do you mean?

mjsheikh:
I was referring to storing the vector (physics context) as an array with different entries representing various dimensions e.g. first one is x etc. 
Or there are better ways for the same?

ChrisRackauckas:
Treating the space as a 3-tensor is probably right.

mjsheikh:
Did some clean-up from last commit, added divergence. So, this is the overall structure.
If its fine then next things in the bucket - 
- [ ] Compatibility with  BC-padded array
- [ ] Few methods for combination and composition of operators
- [ ] Docs update

mjsheikh:
@ChrisRackauckas  So, I think this is comprehensive. Please review. 

Seems the support for BC-padded array would need to wait a bit, atleast till we figure out how to actually do the padding :p

ChrisRackauckas:
Is it still WIP?

docs/src/operators/vector_calculus_operators.md
```diff
@@ -0,0 +1,63 @@
...
+#### Functions
+
+Some common functions used in Vector calculus have been provided :
+
+```julia
+dot_product(A::AbstractArray{Array{T,1},N},B::AbstractArray{Array{T,1},N})
+dot_product!(u::AbstractArray{T,N}, A::AbstractArray{Array{T,1},N},B::AbstractArray{Array{T,1},N})
+
+cross_product(A::AbstractArray{Array{T,1},3},B::AbstractArray{Array{T,1},3})
+cross_product!(u::AbstractArray{Array{T,1},3},A::AbstractArray{Array{T,1},3},B::AbstractArray{Array{T,1},3})
```
ChrisRackauckas, line 56: We might want to let people know they can do:

```julia
const ⋅ = dot_product # \cdot
A ⋅ B

const × = cross_product # \times
A × B
```

src/derivative_operators/vector_algebraic_operations.jl
```diff
@@ -0,0 +1,63 @@
+
+function dot_product(A::AbstractArray{Array{T,1},N},B::AbstractArray{Array{T,1},N}) where {T<:Real,N}
```
ChrisRackauckas, line 2: why AbstractArray{Array}?

Also, does LinearAlgebra.dot not handle this case?

src/derivative_operators/vector_calculus_convolutions.jl
```diff
@@ -0,0 +1,343 @@
...
+# Divergence and Gradient convolutions
+##################################################################################
+
+function LinearAlgebra.mul!(x_temp::AbstractVector, has_vector::Bool, A::DerivativeOperator,x::AbstractVector; overwrite = false)
+    convolve_BC_left!(x_temp, A, x, overwrite = overwrite)
+    convolve_interior!(x_temp, A, x, overwrite = overwrite)
+    convolve_BC_right!(x_temp, A, x, overwrite = overwrite)
+end
+
+function convolve_interior!(x_temp::AbstractVector{T1},  A::DerivativeOperator{T3,N,false}, x::AbstractVector{T2}; overwrite = true) where {T1, T2, T3, N}
```
ChrisRackauckas, line 55: This dispatch already exists?

ChrisRackauckas:
Array{Array} for this is going to be too absurdly slow to be useful. Using at least StaticArrays for the interior would at least be required, but we shouldn't require them from the user. From a 3D matrix `A`, we should be the ones to handle the convolution operations of Curl, Grad, etc. directly instead of requiring the user to change the input format specifically for these operations.

src/derivative_operators/vector_algebraic_operations.jl
mjsheikh: - `AbstractArray{Array}` is the structure I was referring to at the start.
   Each Matrix entry would look like `[f₁ f₂ f₃ ... ]` representing various components of the vector.

- Yeah, just checked in LinearAlgebra. So dot, cross & norm from there will work too, when used as element-wise 
  operations. Should I create these as wrappers for feasibility of use? Also, LinearAlgebra didn't have inplace operations 
  for these, maybe introduce them here as new functions?

src/derivative_operators/vector_calculus_convolutions.jl
mjsheikh: I needed a way to divert the convolutions for vcops from those defined for derivate operators since both of them accept similar arguments but behave really differently. Thats why this dispatch (the function at the start), it has an additional dummy argument and will lead to the convolutions specifically meant for vcops.

docs/src/operators/vector_calculus_operators.md
mjsheikh: Alright

mjsheikh:
> Array{Array} for this is going to be too absurdly slow to be useful. Using at least StaticArrays for the interior would at least be required, but we shouldn't require them from the user. From a 3D matrix `A`, we should be the ones to handle the convolution operations of Curl, Grad, etc. directly instead of requiring the user to change the input format specifically for these operations.

@ChrisRackauckas If you remember I asked this thing on the channel few months back!!
The thing is mutability, I was trying to use `MArray` but that implicitly points to the same reference, so during convolutions everything was getting overwritten!

mjsheikh:
> Array{Array} for this is going to be too absurdly slow to be useful. Using at least StaticArrays for the interior would at least be required, but we shouldn't require them from the user. From a 3D matrix `A`, we should be the ones to handle the convolution operations of Curl, Grad, etc. directly instead of requiring the user to change the input format specifically for these operations.

Sure, will look into this.

mjsheikh:
@ChrisRackauckas 

So this is where the multiplication is taking place in convolutions. This is element-wise.    **(1)**
```
for idx in 1:(A.ops[1].stencil_length)
    x[2] += -cur_coeff_1 * cur_stencil_1[idx] * u[I+e[1]*(idx-mid)][3]
    x[3] += cur_coeff_1 * cur_stencil_1[idx] * u[I+e[1]*(idx-mid)][2]
end
```
I can convert that to : **(2)**
```
idx = convert(SVector{A.ops[1].stencil_length, Int}, 1:A.ops[1].stencil_length)  
temp1 = convert(SVector{length(idx), T}, [u[I+i*e[1]][3] for i in (idx .- mid)])
temp2 = convert(SVector{length(idx), T}, [u[I+i*e[1]][2] for i in (idx .- mid)])
x[2] = sum(-cur_coeff_1 * cur_stencil_1[idx] .* (temp1))
x[3] = sum(cur_coeff_1 * cur_stencil_1[idx] .* (temp2))
```
Is this what you meant?

One more thing that can be done is converting the input `Array{Array}` 3D matrix `A` to `Array{SVector}`, there too we would approach the multiplication as either **(1)** or **(2)**

ChrisRackauckas:
Yes, though maybe keeping it in pure indexing form may make it more amendable to nice transformations via `@turbo` (@chriselrod does the macro work on something that transforms the third dimension into an SVector to then do an operation, or would that ruin its blocking schemes?)

chriselrod:
> Yes, though maybe keeping it in pure indexing form may make it more amendable to nice transformations via `@turbo` (@chriselrod does the macro work on something that transforms the third dimension into an SVector to then do an operation, or would that ruin its blocking schemes?)

Until LoopVectorization gets a front end that can introspect/understand things like that, it'd either not work or ruin the blocking schemes.

mjsheikh:
@ChrisRackauckas These are the benchmarks for `A*u` with `A` = `Curl` (this one is the most complex of all) :

1. For `Array{Array}` and element-wise convolution
```
  memory estimate:  746.22 KiB
  allocs estimate:  14804
  --------------
  minimum time:     468.595 μs (0.00% GC)
  median time:      544.707 μs (0.00% GC)
  mean time:        616.459 μs (11.55% GC)
  maximum time:     8.308 ms (92.88% GC)
  --------------
  samples:          8063
  evals/sample:     1
```
  
2. Converting to `Array{SVector}` and element-wise convolution
  ```
  memory estimate:  792.48 KiB
  allocs estimate:  15550
  --------------
  minimum time:     452.631 μs (0.00% GC)
  median time:      524.720 μs (0.00% GC)
  mean time:        601.082 μs (12.54% GC)
  maximum time:     8.425 ms (93.55% GC)
  --------------
  samples:          8267
  evals/sample:     1
```

3. For `Array{Array}` and `SVector` multiplication
  ```
  memory estimate:  3.58 MiB
  allocs estimate:  62731
  --------------
  minimum time:     4.880 ms (0.00% GC)
  median time:      5.330 ms (0.00% GC)
  mean time:        6.094 ms (10.65% GC)
  maximum time:     20.668 ms (68.10% GC)
  --------------
  samples:          820
  evals/sample:     1
```
  
4. Converting to `Array{SVector}` and `SVector` multiplication 
  ```
  memory estimate:  3.63 MiB
  allocs estimate:  63477
  --------------
  minimum time:     4.811 ms (0.00% GC)
  median time:      5.252 ms (0.00% GC)
  mean time:        5.998 ms (10.68% GC)
  maximum time:     19.189 ms (68.06% GC)
  --------------
  samples:          833
  evals/sample:     1 
```

**Comment by ChrisRackauckas** [comment]
All of these are pretty slow. Why are the allocations so high?

**Comment by mjsheikh** [comment]
> All of these are pretty slow. Why are the allocations so high?

The allocations seem high because we are operating in arrays.
Hence, in convolutions operations like :
 
```
for I in R1
    x = zeros(T,3)
    for idx in 1:(A.ops[3].stencil_length)
        x[1] += -cur_coeff_3*cur_stencil_3[idx]*u[I+e[3]*(idx-mid)][2]
        x[2] += cur_coeff_3*cur_stencil_3[idx]*u[I+e[3]*(idx-mid)][1]
    end
    x_temp[I] += x 
end
```
this is something which is O(N). N = 9x9x9 = 729.
The first case is the fastest.

**Comment by ChrisRackauckas** [comment]
>The allocations seem high because we are operating in multiDim arrays.

That's not true.

>x = zeros(T,3)

That allocation is not necessary. If you do straight indexing and drop that it'll be a lot faster. Take a look at Lecture 2 of https://github.com/mitmath/18337

**Comment by mjsheikh** [comment]
@ChrisRackauckas 
Yes, I removed that allocation. Thanks for pointing out.
But the real culprit was - accessing as `u[I+e[3]*(idx-mid)]`. This was exponentially increasing the allocations. Changed that to `u[I+CartesianIndex(0,0,(idx-mid))]`
Now its down to about 1385 allocations (from 14804!) and those are for allocating the output :
```
 for I in CartesianIndices(x_temp)
        x_temp[I] = zeros(T,3)
 end
```
This is good?

**Comment by ChrisRackauckas** [comment]
Removing `x_temp[I] = zeros(T,3)` is precisely the reason to not use the array of array form.

File: docs/src/operators/vector_calculus_operators.md (hunk 1 above)
Comment (mjsheikh) [reply docs/src/operators/vector_calculus_operators.md:56]: I think this will contradict with other methods?

**Comment by mjsheikh** [comment]
Reduced allocations -
```
 @btime u = A*u0
  40.521 μs (15 allocations: 8.94 KiB)
```

**Comment by ChrisRackauckas** [comment]
Nice! That looks about in the right range. I'll take a closer look later today.

File: src/derivative_operators/vector_algebraic_operations.jl [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+    
+    (size(A) === size(B)) || throw(ArgumentError("Vectors must have the same shape"))
+    (N === N2-1) || throw(ArgumentError("Output should be scalar matrix, one dimension less than input"))
+    
+    for I in CartesianIndices(u)
+        u[I] =  A[I,1]*B[I,1]
+        for i in 2:N
+            u[I] += A[I,i]*B[I,i]
+        end
+    end
```
ChrisRackauckas, line 21: We can do this in a separate PR. But this should flip the loop ordering.

src/derivative_operators/vector_algebraic_operations.jl
```diff
@@ -0,0 +1,54 @@
...
+
+function cross_product!(u::AbstractArray{T1,4},A::AbstractArray{T2,4},B::AbstractArray{T3,4}) where {T1<:Number,T2<:Number,T3<:Number}
+    
+    (size(A) === size(B) && size(A) === size(u))|| throw(ArgumentError("Vectors must have the same shape"))
+    s = size(u)
+    for p in 1:s[1],q in 1:s[2],r in 1:s[3]
+        for i in 1:3
+            u[p,q,r,i] = A[p,q,r,i%3 + 1]*B[p,q,r,(i+1)%3 + 1] - A[p,q,r,(i+1)%3 + 1]*B[p,q,r,i%3 + 1]
+        end
+    end
```
ChrisRackauckas, line 42: Loop ordering

src/derivative_operators/vector_calculus_convolutions.jl
```diff
@@ -0,0 +1,276 @@
...
+    end
+
+    # Along 2nd axis
+    for i in 1:s[1] , j in 1+bpc : s[2]-bpc , k in 1:s[3]
+        cur_stencil_2 = eltype(stencil_2) <: AbstractArray ? stencil_2[j-bpc] : stencil_2
+        cur_coeff_2   = typeof(coeff_2)   <: AbstractArray ? coeff_2[j] : coeff_2 isa Number ? coeff_2 : true
+        for idx in 1:(A.ops[2].stencil_length)
+            x_temp[i,j,k,1] += cur_coeff_2*cur_stencil_2[idx]*u[i,j+idx-mid,k,3]
+            x_temp[i,j,k,3] += -cur_coeff_2*cur_stencil_2[idx]*u[i,j+idx-mid,k,1]
+        end
```
ChrisRackauckas, line 179: loop ordering

ChrisRackauckas:
Let's do a different PR that passes through and correct the loop orders.

PR #376, opened 2021-04-19
Add operator combination with AbstractMatrix and UniformScaling

briochemc:
This PR adds functionality to add `AbstractMatrix` and `UniformScaling` objects to linear operators.

E.g., with this PR you can do
```julia
A = rand(2,2)
B = rand(2,2)
OP = DiffEqArrayOperator(A)
L = OP + B # converts B into a DiffEqArrayOperator before combining them
L = OP + 2I # converts 2I into a diagonal matrix and then applies the rule above
OP = DiffEqArrayOperator(sparse(A))
L = OP + 2I # Also works for sparse operators!
```

I added some dummy tests, but I'm not sure what else I should do. Please let me know!


briochemc:
Thank @ChrisRackauckas! Sorry, I forgot to add a version bump before you merged it...

PR #379, opened 2021-04-21
add grid_align option

### LINKED ISSUE #378: Add option to shift grid by half a point in `MOLFiniteDifference` for conservative method
To allow a `MOLFiniteDifference` method to be conservative, we just need to shift the grid by `dx/2` so that `x=0` and `x=L` are aligned with edges of the grid instead of cell centers. In this case the grid would be `dx/2:dx:L-dx/2` instead of `0:dx:L`.

For example this could be a keyword-argument in `MOLFiniteDifference`: `grid_align="center"` (default, current implementation), or `"edge"`.

In terms of implementation, the only difference would be in the boundary conditions:
- `Dt(u1) ~ Dx((u2-u1)/dx - a)` where `a` is the flux at the left-hand boundary
- `Dt(un) ~ Dx(b - (un-u(n-1))/dx)` where `b` is the flux at the right-hand boundary.

If the boundary conditions are Neumann `a` and `b` are given directly. If the boundary conditions are Dirichlet `a` and `b` can be calculated by extrapolating to ghost nodes. 
Therefore, in the general case 
```julia
f(u(t,0),Dx(u(t,0)) ~ c
```
we can replace `Dx(u(t,0))` with `a` and `u(t,0)` with `u1-a*dx/2`, which then gives a bc equation for the flux `a` (and similarly on the right-hand boundary).

(This is like a Finite Volume discretization, but on a regular grid the result is the same)

Will wait for discussions in this issue before trying to implement this

tinosulzer:
Fixes #378 
Adds new tests for conservation of mass with Neumann BCs


src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -5,14 +5,15 @@ struct MOLFiniteDifference{T,T2} <: DiffEqBase.AbstractDiscretization
     time::T2
     upwind_order::Int
     centered_order::Int
+    grid_align::String
```
ChrisRackauckas, line 8: We should use enums here. `@enum GridAlign EdgeAlign CenterAlign`. That would naturally catch any syntax issues.

ChrisRackauckas:
Changing to enums I think is the only qualm here.

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -1,18 +1,22 @@
 using ModelingToolkit: operation, istree, arguments
 # Method of lines discretization scheme
+
+@enum GridAlign center_align edge_align
+center_align
```
ChrisRackauckas, line 5:
```suggestion
```

ChrisRackauckas (approved):

Issue #282, opened 2020-10-02
MOLFiniteDifference: NonlinearProblem


ChrisRackauckas:
https://github.com/SciML/DiffEqOperators.jl/pull/250#discussion_r498583409

Basically, I think it needs to be like `MOLFiniteDifference(dx,odedim = t)`, i.e., you need to choose which dimension to not discretize. If you discretize all dimensions, then you get a NonlinearProblem, i.e. a nonlinear system that is implicit in all directions. This would allow us to handle things in a more principled way rather than relying on `t`.

ChrisRackauckas:
https://github.com/SciML/DiffEqOperators.jl/pull/349 fixes the preferred dimension by making the choice required. However, we should have a `ImplicitFiniteDifference` discretization for handling the case of `NonlinearProblem`.

ChrisRackauckas:
Yes exactly.

ChrisRackauckas:
This was completed.

PR #349, opened 2021-03-27
New MOLFiniteDifference Discretization

ChrisRackauckas:
```julia
using ModelingToolkit, DiffEqOperators, LinearAlgebra

# 3D PDE
@parameters t x y
@variables u(..)
Dxx = Differential(x)^2
Dyy = Differential(y)^2
Dt = Differential(t)
t_min= 0.
t_max = 2.0
x_min = 0.
x_max = 2.
y_min = 0.
y_max = 2.

# 3D PDE
eq  = Dt(u(t,x,y)) ~ Dxx(u(t,x,y)) + Dyy(u(t,x,y))

analytic_sol_func(t,x,y) = exp(x+y)*cos(x+y+4t)
# Initial and boundary conditions
bcs = [u(t_min,x,y) ~ analytic_sol_func(t_min,x,y),
       u(t,x_min,y) ~ analytic_sol_func(t,x_min,y),
       u(t,x_max,y) ~ analytic_sol_func(t,x_max,y),
       u(t,x,y_min) ~ analytic_sol_func(t,x,y_min),
       u(t,x,y_max) ~ analytic_sol_func(t,x,y_max)]

# Space and time domains
domains = [t ∈ IntervalDomain(t_min,t_max),
           x ∈ IntervalDomain(x_min,x_max),
           y ∈ IntervalDomain(y_min,y_max)]
pdesys = PDESystem([eq],bcs,domains,[t,x,y],[u(t,x,y)])

# Method of lines discretization
dx = 0.1; dy = 0.2
discretization = MOLFiniteDifference([x=>dx,y=>dy],t)
prob = ModelingToolkit.discretize(pdesys,discretization)

using OrdinaryDiffEq
sol = solve(prob,Tsit5())
using Plots
plot(sol)
savefig("plot.png")
```

PR #400, opened 2021-05-24
creating nonlinear problem with ImplicitFiniteDifference

akashkgarg:
An attempt to resolve Issue #282 . This introduces a new discretization type, `ImplicitFiniteDifference` that is then used to create a `NonlinearProblem` for equations of the form: `0 ~ Dxx + Dyy  + f(x,y)`, without the dependence of a time variable.


ChrisRackauckas:
I think this PR needs to try and make use of as much of the MOLFiniteDifference code as possible, otherwise it might be for the most part a duplicate. Exactly how to do that might be a little tricky though. I was thinking it would happen in the same function and just try to remove the ODE-specific parts. Thoughts @tinosulzer ?

akashkgarg:
I definitely agree. I don’t intend to duplicate the code as I’ve done here in the final iteration. I duplicated the code here so I can test both branches, making sure I don’t break anything and work through how to refactor the common pieces. Comments/ideas definitely welcome

tinosulzer:
I don't understand why this needs a separate method `ImplicitFiniteDifference`. `MOLFiniteDifference` can already deal with PDAEs, see https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_1D_PDAE.jl. So don't we just need to recognize if there are no time derivatives in the discretized problem and return a `NonlinearSystem` instead of `ODESystem` in this line if that is the case? https://github.com/SciML/DiffEqOperators.jl/blob/b239832226c087e1ebb8f32d37cd2b5f42209386/src/MOLFiniteDifference/MOL_discretization.jl#L345

akashkgarg:
@tinosulzer I believe that #349 makes the time variable mandatory. I was going off of Chris's comment here: https://github.com/SciML/DiffEqOperators.jl/issues/282#issuecomment-808881861. Are you suggesting something like making the time variable optional or perhaps setting it to `nothing`, detect if this is the case and create a `NonlinearProblem` instead? I think one intent was to create orthogonal types for something that depends on space + time (`MOLFiniteDifference`) vs. just space (`ImplicitFiniteDifference`) -- although if that the major difference, maybe we need different names? 

I also tried refactoring this a little bit and it does seem that `NonlinearSystem` has a very different structure from `PDESystem`, so getting those to work interchangeably might be challenge; for example, `NonlinearSystem` does not have something called `indvars`. If a `PDESystem` can be fed into a `NonlinearProblem`, that might be an easier approach.

ChrisRackauckas:
>I also tried refactoring this a little bit and it does seem that NonlinearSystem has a very different structure from PDESystem, so getting those to work interchangeably might be challenge; for example, NonlinearSystem does not have something called indvars. If a PDESystem can be fed into a NonlinearProblem, that might be an easier approach.

You could make it a SteadyStateProblem which upconverts to a NonlinearProblem, where one axis is randomly chosen to be time, but that seems like a very odd way to do it. I think it would be easier to just have an if statement around whether there's variable dependence in the definition of the states, and an if statement on the portion that finds the initial conditions. Then an if on whether to generate a NonlinearSystem or ODESystem. To me it sounds like that wouldn't be too difficult of a way to support the NonlinearSystem in the same function, but I haven't tried doing it.

tinosulzer:
> I think it would be easier to just have an if statement around whether there's variable dependence in the definition of the states, and an if statement on the portion that finds the initial conditions. Then an if on whether to generate a NonlinearSystem or ODESystem. To me it sounds like that wouldn't be too difficult of a way to support the NonlinearSystem in the same function, but I haven't tried doing it.

Yes that is exactly what I was thinking. In terms of user interface, either we could make the user call `MOLFiniteDifference(dxs,nothing)` or have `ImplicitFiniteDifference(dxs)` just to wrap that call. Then the discretization sees if `t` is a variable or `nothing` and chooses those if statements accordingly.

akashkgarg:
This sounds like a good plan. I'll work along that direction. I updated the code based on your suggestions, it's still not quite working yet as I need to work out how to do the substitution rules properly when `t` is not present, I think certain BC rules assume a dependence on `t`.

akashkgarg:
I've almost got something working except getting some very strange behavior with the structural simplify step. I create a NonlinearSystem, which gives me the following systems of equations: 
```julia
11-element Vector{Equation}:
 100.0u₁ + 99.99999999999999u₃ - (200.0u₂) ~ 0
 100.00000000000001u₂ + 100.00000000000003u₄ - (200.00000000000006u₃) ~ 0
 100.00000000000001u₃ + 99.99999999999996u₅ - (199.99999999999997u₄) ~ 0
 99.99999999999996u₄ + 100.00000000000001u₆ - (199.99999999999997u₅) ~ 0
 100.00000000000004u₅ + 100.00000000000006u₇ - (200.0000000000001u₆) ~ 0
 100.00000000000004u₆ + 100.00000000000006u₈ - (200.0000000000001u₇) ~ 0
 99.99999999999999u₇ + 99.99999999999987u₉ - (199.99999999999986u₈) ~ 0
 99.99999999999999u₁₀ + 99.99999999999987u₈ - (199.99999999999986u₉) ~ 0
 100.00000000000006u₁₁ + 100.00000000000004u₉ - (200.0000000000001u₁₀) ~ 0
 u₁ ~ 1
 u₁₁ ~ 1
```

After structural simplification, I get the following: 
```julia
1-element Vector{Equation}:
 0 ~ 500.00000000000364u₃ - 500.0000000000009
```

akashkgarg:
Ignoring the structural simplification bug for the time being it seems that NonlinearSystem does have a problem with solving this very simple linear system (see "test/MOL_NonlinearProblem.jl"). The solution is NaN and the Jacobian is singular. I think the Jacobian in this case should be the Laplacian matrix. Not sure if it's something to do with how I have setup the problem. Perhaps instead of NonlinearSystem we need a LinearSystem to solve equations that arise from finite differences?

ChrisRackauckas:
>Perhaps instead of NonlinearSystem we need a LinearSystem to solve equations that arise from finite differences?

No, the Jacobian would only be singular if the linear system was singular, which wouldn't solve anyways. Newton's method on a linear `f` converges in one step, so it's not even inefficient if the Jacobian is calculated directly. It's worth finding out why it's doing this: my guess is that it's a boundary condition problem. Double check that the generated equations give the right matrix with `u = one(u)` is passed in?

akashkgarg:
Yes, the plot thickens. It seems that the generated function always returns a constant value irrespective of the `u` that I pass into it...

akashkgarg:
Diving a bit deeper looking at the generated symbolics code and jacobians, it seems that I was setting up the problem incorrectly. Instead of something like this: 
```julia
    eqs = [a ~ 1,
           a + c - 2*b ~ 0,
           b + d - 2*c ~ 0,
           d ~ 1]
```

The equations need to be setup like this: 
```julia
    eqs = [0 ~ a - 1,
           0 ~ a + c - 2*b,
           0 ~ b + d - 2*c,
           0 ~ d - 1]
```
requiring the lhs to be zero and the rhs as the expression of all variables. This simple hand-rolled example works when fed into the `NonlinearProblem`. I'll make this change inside the discretization step.

Do we want to support the equivalent formulations of these equations? For example, the following compiles to incorrect code in Symbolics:
```julia
    eqs = [a ~ 1,
           a + c - 2*b ~ 0,
           b + d - 2*c ~ 0,
           d ~ 1]
```
and 
```julia
    eqs = [1 ~ a,
           0 ~ a + c - 2*b,
           0 ~ b + d - 2*c,
           1 ~ d]
```

ChrisRackauckas:
Yeah that's an upstream MTK issue. It should move everything over at lowering time. That's worth an issue.

ChrisRackauckas:
`eqs = 0 .~ eqs.rhs .- eqs.lhs` needs to happen in the lowering.

akashkgarg:
Moving this back to draft to add documentation of this new feature and add at least one more test for the 2D case.

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -179,7 +188,22 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
...
+                            bcargs = arguments(first(depvarbcmaps[i]))
+                            # Replace Differential terms in the bc lhs with the symbolic spatially discretized terms
+                            # TODO: Fix Neumann and Robin on higher dimension
+                            lhs = nspace == 1 ? substitute(bc.lhs,depvarderivbcmaps[i]) : bc.lhs
+
+                            # Replace symbol in the bc lhs with the spatial discretized term
+                            lhs = substitute(lhs,depvarbcmaps[i])
+                            rhs = bc.rhs
+                            lhs = lhs isa Vector ? lhs : [lhs] # handle 1D
+                            push!(bceqs,lhs .~ rhs)
```
tinosulzer, line 204: seems like a lot of code duplication here

tinosulzer (commented):
Looks on the right track

tinosulzer:
> After structural simplification, I get the following:
```
1-element Vector{Equation}:
 0 ~ 500.00000000000364u₃ - 500.0000000000009
```

Coming back to this, it seems `structural_simplify` is basically able to solve this system symbolically but isn't taking the last step so that there is still one equation to "solve" - is that correct? In that case could there be some kind of `DummyNonlinearProblem` for this case that acts like a `NonlinearProblem` but just returns the known symbolic solution?

ChrisRackauckas:
Yeah, ODEs have the same thing where if you write a fully explicit equation with no actual differential variables something weird happens. It's an edge case, we know, but we just haven't made the edge case have a nicer solution. @YingboMa they noticed.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "4.27.0"
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 BlockBandedMatrices = "ffab5731-97b5-5995-9138-79e8c1846df0"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
YingboMa: Oof. Wouldn't this be circular dependency? Why do you need DifferentialEquations?

Project.toml
```diff
@@ -14,6 +14,7 @@ LazyBandedMatrices = "d7e5e226-e90b-4449-9968-0f923699bf6f"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
+NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
```
YingboMa, line 17: This need a compat.

ChrisRackauckas (approved):
Wow, this came out rather clean, great!

tinosulzer:
Nice!

PR #403, opened 2021-05-31
Specify type for dx in difference constructors

JordiBolibar:
Only the type of `dx` was missing in the DiffEqOperators.jl documentation, which made it unclear if `dx` needs to be an `Int` or a `Float`.

### DIFF
--- a/docs/src/operators/derivative_operators.md
+++ b/docs/src/operators/derivative_operators.md
@@ -91,11 +91,11 @@ The constructors are as follows:
 
 ```julia
 CenteredDifference{N}(derivative_order::Int,
-                      approximation_order::Int, dx,
+                      approximation_order::Int, dx::Float,
                       len::Int, coeff_func=nothing)
 
 UpwindDifference{N}(derivative_order::Int,
-                    approximation_order::Int, dx
+                    approximation_order::Int, dx::Float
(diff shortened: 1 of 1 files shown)


docs/src/operators/derivative_operators.md
```diff
@@ -91,11 +91,11 @@ The constructors are as follows:
 
 ```julia
 CenteredDifference{N}(derivative_order::Int,
-                      approximation_order::Int, dx,
+                      approximation_order::Int, dx::Float,
```
ChrisRackauckas, line 94: That doesn't necessarily need to be the case though. It depends on the values and such, and even then it would be `AbstractFloat`. But that's too restrictive to be true.

JordiBolibar: OK, I see. It's just that the error code is pretty misleading. It would be pretty easy to spot with a clearer message.

ChrisRackauckas: I agree the error message could be better, but I think we need to also enforce correctness.

JordiBolibar: Right, I'll close this one and I'll make another one with a meaningful comment in order to avoid confusion regarding this.

PR #412, opened 2021-06-23
Fixing coefficient functions in CenteredDifference

### LINKED ISSUE #411: CenteredDifference gives zero matrix with constant coeff_func
when we have 
`x = collect(0:0.01:1)`
these two give different result: 
`CenteredDifference(2,2,0.01, length(x))*RobinBC((1.0, -1.0, 1.0), (0., 1.,0.), 0.01, length(x))*exp.(x)`
` CenteredDifference(2,2,0.01, length(x), 1)*RobinBC((1.0, -1.0, 1.0), (0., 1.,0.), 0.01, length(x))*exp.(x)`
the only difference is only coefficient function. 
actually, no matter what the coeff_fun is, CenteredDifference gives a zero matrix
`Array(CenteredDifference(2,2,0.01, length(x), constant))`

toollu:
Should fix #411 

I don't know the package in detail but at a first glance the CenteredDifference constructors miss the call to `compute_coeffs`
if `coeffs != nothing` .

### DIFF
--- a/src/derivative_operators/derivative_operator.jl
+++ b/src/derivative_operators/derivative_operator.jl
@@ -90,6 +90,12 @@ function CenteredDifference{N}(derivative_order::Int,
     offside = 0
 
     coefficients            = coeff_func isa Nothing ? nothing : fill!(Vector{T}(undef,len),0)
+    
+    if coeff_func != nothing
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Thanks!

PR #414, opened 2021-06-25
Multi-Dimensional BC Operator

mjsheikh:
Here, the last implementation has been edited to produce the expected results.

There was a goof-up in previous structure since it presumed that all data columns along a dimension would have same boundary conditions. I have rather used extrapolation using FD approximation to estimate the right BVs.
Moreover, a multi-dimensional boundary padding in true-sense would require updation of 'unpadded' part of our `Array` after each operation along a dimension. Hence, handling it that way has been administered.
 
 There are still few bottlenecks that need to be thought about and resolved if feasible -
- [ ] Producing concretizations of the operator & its composite
- [ ] The output returned is an `Array` rather than `CompositeBoundaryPaddedArray` because segregating what's padded and what's not becomes cumbersome as the BVs after first operation are interior points for the next one.

### DIFF
--- a/src/derivative_operators/concretization.jl
+++ b/src/derivative_operators/concretization.jl
@@ -163,190 +163,190 @@ function Base.convert(::Type{AbstractMatrix},A::AbstractBC{T}) where T
 end
 
 # Multidimensional BC operators
-_concretize(Q::MultiDimDirectionalBC, M) = _concretize(Q.BCs, M)
+# _concretize(Q::MultiDimDirectionalBC, M) = _concretize(Q.BCs, M)
 
 function _concretize(Q::AbstractArray{T,N}, M) where {T,N}
     return (stencil.(Q, fill(M,size(Q))), affine.(Q))
 end
 
-function LinearAlgebra.Array(Q::MultiDimDirectionalBC{T, B, D, N, L}, s::NTuple{N,G}) where {T, B, D,N,L, G<:Int}
-    @assert size(Q.BCs) == perpindex(s, D) "The size of the BC array in Q, $(size(Q.BCs)) is incompatible with s, $s"
-    blip = zeros(Int64, N)
-    blip[D] = 2
-    s_pad = s.+ blip # extend s in the right direction
-    Q = _concretize(Q.BCs, s[D])
-    ē = unit_indices(N)
-    QL = zeros(T, prod(s_pad), prod(s))
-    Qb = zeros(T, prod(s_pad))
-    ranges = Union{typeof(1:10), Int64}[1:s[i] for i in 1:N]
-    ranges[D] = ranges[D] .+ 1
+# function LinearAlgebra.Array(Q::MultiDimDirectionalBC{T, B, D, N, L}, s::NTuple{N,G}) where {T, B, D,N,L, G<:Int}
+#     @assert size(Q.BCs) == perpindex(s, D) "The size of the BC array in Q, $(size(Q.BCs)) is incompatible with s, $s"
+#     blip = zeros(Int64, N)
+#     blip[D] = 2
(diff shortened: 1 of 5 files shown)


src/derivative_operators/concretization.jl
```diff
@@ -163,190 +163,190 @@ function Base.convert(::Type{AbstractMatrix},A::AbstractBC{T}) where T
 end
 
 # Multidimensional BC operators
-_concretize(Q::MultiDimDirectionalBC, M) = _concretize(Q.BCs, M)
+# _concretize(Q::MultiDimDirectionalBC, M) = _concretize(Q.BCs, M)
```
ChrisRackauckas, line 166: what's the reason for commenting instead of removing?

src/derivative_operators/multi_dim_bc_operators.jl
```diff
@@ -1,8 +1,26 @@
...
+    stencils::S
+end
+
+struct MultiDimBC{N} end
+
+MultiDimBC{dim}(N::Int, approx_order::Int, dx::T) where {dim,T} = finite_stencil(approx_order,dx,dim,N)
+
+"""
+slice_rmul lets you multiply each vector like strip of an array `u` with a linear operator `A`, sliced along dimension `dim`
+"""
```
ChrisRackauckas, line 20: no space or this won't be a docstring.

src/derivative_operators/multi_dim_bc_operators.jl
```diff
@@ -1,8 +1,26 @@
...
+    r :: T
+end
+
+struct ComposedMultiDimBC{T, S, N, M} <: MultiDimensionalBC{T, N}
+    stencils::S
+end
+
+struct MultiDimBC{N} end
+
+MultiDimBC{dim}(N::Int, approx_order::Int, dx::T) where {dim,T} = finite_stencil(approx_order,dx,dim,N)
```
ChrisRackauckas, line 16: what's N?

test/DerivativeOperators/multi_dim_bc_test.jl
```diff
@@ -1,104 +1,108 @@
...
+	################################################################################
+	# Test 2d extension
+	################################################################################
+
+	s = x, y = (-1.9:0.1:1.9, -1.9:0.1:1.9)                 # domain of unpadded function
+	dx = dy = x[2] - x[1]
+	paraboloid(x::T, y::T) where T = 2*(x^2+y^2) - 4        # declare an elliptic paraboloid function
+	u0 = [paraboloid(X, Y) for X in x, Y in y]
+
+	q1 = MultiDimBC{1}(2,6,dx)                              # Padding along x
```
ChrisRackauckas, line 13: I don't think this can account for affine terms or Dirichlet non-zero?

src/derivative_operators/concretization.jl
mjsheikh: Earlier it was an Affine operator with concretizations. Now it isn't and hence won't support this but rather than removing I thought about commenting out keeping in mind that if by any means one can come up with a way to still concretize this later. 
Now when I think of it, I don't know why this should necessarily be manipulated and concretized  ┐(•_•)┌

src/derivative_operators/multi_dim_bc_operators.jl
mjsheikh: Overall Dimension of the space in which its operating. I think the declaration in the struct `MultiDimBC{N}` is misleading.
Will replace there with `dim`.

test/DerivativeOperators/multi_dim_bc_test.jl
mjsheikh: Yeah, this isn't an affine operator per se.

PR #424, opened 2021-07-07
Added link to SciMLBase

killah-t-cell:


ChrisRackauckas:
We shouldn't point to the code there. We should really get a page here:

https://scimlbase.sciml.ai/dev/

I think https://diffeq.sciml.ai/stable/features/diffeq_operator/ is the most complete page right now.

killah-t-cell:
Sounds good. I changed the link to https://diffeq.sciml.ai/stable/features/diffeq_operator/ for now.

ChrisRackauckas:
Thanks! BTW since you're a sponsor, I'm hoping to use the sponsor funds to soon setup a bounty to get more of these fundamental interfaces up and documented, add more docstrings, etc. This is probably our biggest problem right now.

killah-t-cell:
Cool. Glad I can help. Also, are there any interfaces that would be particularly educational to document?

ChrisRackauckas:
http://nonlinearsolve.sciml.ai/dev/ and https://galacticoptim.sciml.ai/dev/ are pretty bare. https://github.com/SciML/LinearSolvers.jl will need docs setup soon. https://github.com/SciML/Quadrature.jl should get real docs in the standard form. As in the SciMLBase docs, I'm hoping that all of these numerical problems gets a standard numeric and symbolic interface so that way PDE tools can easily transform equations to the different forms and hit the various solvers. We're almost there.

PR #427, opened 2021-07-21
reduced allocations for vcops and fixed loop orders

mjsheikh:
Written new convolutions for reducing allocations in `Gradient` and `Divergence` treating 2D & 3D functions/vectors as special cases since they are the ones most widely used.
Loop ordering has also been fixed wherever required.

### DIFF
--- a/src/derivative_operators/vector_algebraic_operations.jl
+++ b/src/derivative_operators/vector_algebraic_operations.jl
@@ -35,10 +35,8 @@ function cross_product!(u::AbstractArray{T1,4},A::AbstractArray{T2,4},B::Abstrac
     
     (size(A) === size(B) && size(A) === size(u))|| throw(ArgumentError("Vectors must have the same shape"))
     s = size(u)
-    for p in 1:s[1],q in 1:s[2],r in 1:s[3]
-        for i in 1:3
-            u[p,q,r,i] = A[p,q,r,i%3 + 1]*B[p,q,r,(i+1)%3 + 1] - A[p,q,r,(i+1)%3 + 1]*B[p,q,r,i%3 + 1]
-        end
+    for i in 1:3, r in 1:s[3], q in 1:s[2], p in 1:s[1]
+        u[p,q,r,i] = A[p,q,r,i%3 + 1]*B[p,q,r,(i+1)%3 + 1] - A[p,q,r,(i+1)%3 + 1]*B[p,q,r,i%3 + 1]
     end
 end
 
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Share some benchmarks.

mjsheikh:
**Benchmarks** :
- Earlier for generalized implementation of Divergence & Gradient, for a 3D function/Vector
```
Divergence
@btime u = A*u0
  147.637 μs (1859 allocations: 107.80 KiB)

Gradient
@btime u = A*u0
  148.795 μs (1908 allocations: 114.75 KiB)
```
- After specialization, for a 3D function/Vector
```
Divergence
@btime u = A*u0
  31.761 μs (15 allocations: 3.52 KiB)

Gradient
@btime u = A*u0
  30.562 μs (16 allocations: 8.95 KiB)
```
- For Curl, before & after fixing some loop orders
```
@btime u = A*u0
  44.124 μs (15 allocations: 8.94 KiB)

@btime u = A*u0
  42.645 μs (15 allocations: 8.94 KiB)
  ```
  
- This is the part that is performing better by the first implementation (for `dot_product`), hence unchanged :
``` 
for I in CartesianIndices(u)
	u[I] =  A[I,1]*B[I,1]
	for i in 2:N
    		u[I] += A[I,i]*B[I,i]
	end
end
@btime D = dot_product(u0,u1)
  3.173 μs (1 allocation: 5.88 KiB)
```
```
for l in 1:N
        if l==1
            for I in CartesianIndices(u)
                u[I] =  A[I,1]*B[I,1]
            end
        else
            for I in CartesianIndices(u)
                u[I] +=  A[I,l]*B[I,l]
            end
        end
end
@btime D = dot_product(u0,u1)
  4.293 μs (1 allocation: 5.88 KiB)
```

**Comment by ChrisRackauckas** [comment]
>This is the part that is performing better by the first implementation (for dot_product), hence unchanged :

Yes, you want to avoid conditionals in inner loops as much as possible. They will cause branch checks which can stop SIMD from being possible.

File: src/derivative_operators/vector_algebraic_operations.jl [hunk 1]
```diff
@@ -35,10 +35,8 @@ function cross_product!(u::AbstractArray{T1,4},A::AbstractArray{T2,4},B::Abstrac
     
     (size(A) === size(B) && size(A) === size(u))|| throw(ArgumentError("Vectors must have the same shape"))
     s = size(u)
-    for p in 1:s[1],q in 1:s[2],r in 1:s[3]
-        for i in 1:3
-            u[p,q,r,i] = A[p,q,r,i%3 + 1]*B[p,q,r,(i+1)%3 + 1] - A[p,q,r,(i+1)%3 + 1]*B[p,q,r,i%3 + 1]
-        end
+    for i in 1:3, r in 1:s[3], q in 1:s[2], p in 1:s[1]
+        u[p,q,r,i] = A[p,q,r,i%3 + 1]*B[p,q,r,(i+1)%3 + 1] - A[p,q,r,(i+1)%3 + 1]*B[p,q,r,i%3 + 1]
```
ChrisRackauckas, line 39: I think it might be more efficient to hoist the `i%3 + 1` out of the inner loop? Mod operations are expensive. Actually, if you can write that without a mod or an if statement that would be best.

ChrisRackauckas:
Let's merge this as a big improvement. Next, take a look at hoisting out the mod and try adding `LoopVectorization.@turbo` to some of the big loops.

PR #428, opened 2021-07-23
Loop vectorization for major convolutions of vcops

mjsheikh:
Some major enhancements via using @turbo

**Benchmarks** :
- Earlier
```
Divergence
@btime u = A*u0
  31.761 μs (15 allocations: 3.52 KiB)

Gradient
@btime u = A*u0
  30.562 μs (16 allocations: 8.95 KiB)

Curl
@btime u = A*u0
  42.645 μs (15 allocations: 8.94 KiB)
```

- Now
```
Divergence
@btime u = A*u0
  8.255 μs (15 allocations: 3.52 KiB)

Gradient
@btime u = A*u0
  8.933 μs (16 allocations: 8.95 KiB)

Curl
@btime u = A*u0
  11.426 μs (15 allocations: 8.94 KiB)
```
Some other operations have also been optimized (mod removal, @turbo in `dot` , `cross`)


mjsheikh:
**Update** : @turbo would work for uniform grids, since LoopVectorization doesn't allow a SMatrix kind of access in nested loops so that has been skipped at some places in new commit.

@ChrisRackauckas I think if we want to use `LoopVectorization` intensively in future for this repo, then it would be a better idea to use `HybridArrays` rather than `Array{SArray}`. For e.g. stencils for non-uniform grid `DerivateOperator`should be stored as `HybridArrays`.
Perhaps a separate PR should be done for that?

mjsheikh:
Also, I never modified the code whose CI tests are failing here. They are going pretty well on my system. Could you please check? :sweat_smile:

Issue #385, opened 2021-05-05
Allow higher order derivatives in Boundary Conditions


akashkgarg:
Currently the discretization assumes either Dirichlet boundary conditions or 1st order Neumann BCs. There are however PDEs that require higher order BCs; e.g., the beam equation with 3rd order BCs: 
```julia
    @parameters x, t
    @variables u(..)
    Dt = Differential(t)
    Dtt = Differential(t)^2
    Dx = Differential(x)
    Dxx = Differential(x)^2
    Dx3 = Differential(x)^3
    Dx4 = Differential(x)^4

    g = -9.81
    EI = 1
    mu = 1
    L = 10.0
    dx = 0.4

    eq = Dtt(u(t,x)) ~ -mu*EI*Dx4(u(t,x)) + mu*g

    bcs = [u(0, x) ~ 0,
           u(t,0) ~ 0,
           Dx(u(t,0)) ~ 0,
           Dxx(u(t, L)) ~ 0,
           Dx3(u(t, L)) ~ 0]

    # Space and time domains
    domains = [t ∈ IntervalDomain(0.0,1.0),
               x ∈ IntervalDomain(0.0,L)]

    pdesys = PDESystem(eq,bcs,domains,[t,x],[u(t,x)])
    discretization = MOLFiniteDifference([x=>dx],t, centered_order=4)
    prob = discretize(pdesys,discretization)
```

A related issue is also reported here with second order BCs:  
https://github.com/SciML/DiffEqOperators.jl/issues/344

Issue #384, opened 2021-04-29
Better testing


tinosulzer:
Currently most tests are of the form
```julia
@test a ≈ b atol=0.01
```
which uses the L2-norm and so can give errors even if the vectors are equal within tol elementwise (L-inf norm).  (Linf(x) <= L2(x) <= sqrt(n)Linf(x))
Should we be testing instead
```julia
@test all(isapprox.(a, b, rtol = 0.01))
```
? Or do we really want to test the L2 norm?

See https://discourse.julialang.org/t/unit-test-equality-for-arrays/8138/13

ChrisRackauckas:
Element-wise testing is probably better.

tinosulzer:
So using that isapprox? Or is there a cleaner way?

ChrisRackauckas:
Broadcasted isapprox is the right idea.

PR #435, opened 2021-08-11
MOLFiniteDifference : Adding support for higher derivative orders in bcs

mjsheikh:


mjsheikh:
For use cases like #344 , #385 etc. support for N-order derivatives has been provided. 
Since there were many flagged issues in the KS eqn, I have left that as `@test_broken`, though now atleast it doesn't error due to mishandling N-ordered derivatives and produces output.

src/MOLFiniteDifference/MOL_discretization.jl
```diff
@@ -202,7 +211,10 @@ function SciMLBase.symbolic_discretize(pdesys::ModelingToolkit.PDESystem,discret
                             bcargs = arguments(first(depvarbcmaps[i]))
                             # Replace Differential terms in the bc lhs with the symbolic spatially discretized terms
                             # TODO: Fix Neumann and Robin on higher dimension
-                            lhs = nspace == 1 ? substitute(bc.lhs,depvarderivbcmaps[i]) : bc.lhs
+                            # Update: Fixed for 1D systems
```
tinosulzer, line 214: can remove this imo

tinosulzer (commented):
Can you add a test with Neumann BCs to this? https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_2D_Diffusion.jl

src/MOLFiniteDifference/MOL_discretization.jl
mjsheikh: > Can you add a test with Neumann BCs to this? https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_2D_Diffusion.jl

Yeah, so the support is for higher ordered Neumann in 1D systems.
Perhaps a test should be added in :
https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_1D_HigherOrder.jl ?

mjsheikh: > can remove these two comments imo
I think that is still a TODO

mjsheikh:
@tinosulzer tests added.

test/MOL/MOL_1D_Linear_Diffusion.jl
```diff
@@ -59,6 +59,37 @@ using ModelingToolkit: Differential
             @test all(isapprox.(u_approx, exact, atol=0.01))
         end
     end
+    
+    @parameters t x
+    
+    # BCs with 2nd order Neumann
+    bcs = [u(0,x) ~ cos(x),
+           Dxx(u(t,0)) ~ -exp(-t),
+           Dxx(u(t,Float64(π))) ~ exp(-t)]
```
tinosulzer, line 68: You can't specify a 2nd order BC on a 2nd order system. See https://github.com/SciML/DiffEqOperators.jl/pull/382#issuecomment-828782462

test/MOL/MOL_1D_Linear_Diffusion.jl
```diff
@@ -545,6 +577,33 @@ end
...
         @test all(isapprox.(v_exact(x_sol, t_sol[i]), sol.u[i][l-1:end], atol=0.01))
     end
+
+    @parameters t x
+
+    # BCs with 3rd order Neumann
+    bcs = [u(0,x) ~ cos(x),
+           v(0,x) ~ sin(x),
+           u(t,0) ~ exp(-t),
+           Dxxx(u(t,1)) ~ exp(-t) * sin(1),
```
tinosulzer, line 587: BC order should be <= 1

tinosulzer (commented):
Sorry I misunderstood what you were doing. Yes, the tests should be in https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_1D_HigherOrder.jl (ideally just fix the tests that are `test_broken` )

mjsheikh:
> Sorry I misunderstood what you were doing. Yes, the tests should be in https://github.com/SciML/DiffEqOperators.jl/blob/master/test/MOL/MOL_1D_HigherOrder.jl (ideally just fix the tests that are `test_broken` )

Ideally I should have fixed those broken tests but I think some other pieces are missing in handling beam eqns (like those mentioned in  https://github.com/SciML/DiffEqOperators.jl/pull/382#issuecomment-832184151 etc.) :sweat_smile:

test/MOL/MOL_1D_Linear_Diffusion.jl
mjsheikh: Yes, this was just to show that desired operation gets carried out safely and in correct manner. 
But yeah, I think a higher ordered eqn would be good.

mjsheikh:
I think the KdV wave eqn should be a good test case. The no. of BCs have been set more than required though, to produce an accurate solution.

mjsheikh:
@tinosulzer this looks good?

test/MOL/MOL_1D_HigherOrder.jl
```diff
@@ -120,3 +120,51 @@ end
...
+    @test sol.retcode == :Success
+
+    xs = domains[1].domain.lower+dx+dx+dx:dx:domains[1].domain.upper-dx-dx
+    ts = sol.t
+
+    u_predict = sol.u
+    u_real = [[u_analytic(x, t) for x in xs] for t in ts]
+    u_diff = u_real - u_predict
+    for i in 1:length(u_diff)
+    	@test u_diff[i] ≈ zeros(length(u_diff[i])) atol=0.05;
```
tinosulzer, line 168: see #384

tinosulzer (approved):
Looks good, can the other (broken) KS test now be removed? And why do the beam equation tests still not work after this fix?

test/MOL/MOL_1D_HigherOrder.jl
mjsheikh: > Looks good, can the other (broken) KS test now be removed? And why do the beam equation tests still not work after this fix?

Yes, we can actually remove the KS test. I tested that, it would just require more BCs to pass but that seems unnecessary since we already have the KdV test.
As for the beam eqn, thats related to https://github.com/SciML/DiffEqOperators.jl/pull/382#issuecomment-832184151 **point 2**.

PR #473, opened 2021-09-08
Fix spherical diffusion tests

tinosulzer:
- Added `@named` to all systems
- Fixed a few warnings
- Currently, rules for nonlinear and spherical operations don't work (all return `nothing`)


tinosulzer:
lol forgot to pull upstream master

PR #481, opened 2021-10-12
VecJacOperator

avik-pal:


avik-pal:
@ChrisRackauckas how do I prevent it from canceling the other tests if MOL fails?

tinosulzer:
Same error used to happen with MTK 6.4.7
MTK 6.4.9 fixed it
Now MTK 6.5.2 is broken again

ChrisRackauckas:
@shashi can we figure out what happened to that rule application?

PR #501, opened 2021-12-22
fornberg(2020) weights based on hermite-based finite difference

yewalenikhil65:
As per discussion earlier today with @ChrisRackauckas  on slack, this PR extends `DiffEqOperators.calculate_weights` function to include recent algorithm of hermite-based finite difference weights by [fornberg(2020)](https://www.colorado.edu/amath/sites/default/files/attached-files/2020_f_hermite-fd_ima_j_num_anal.pdf)

- The original fornberg(1988) algorithm is invoked by default !
- To invoke hermite-based FD fornberg(2020), one just needs to pass kwarg `dfdx = true` to the `DiffEqOperators.calculate_weights` function, thus, returning the weights/stencil for function values and weights/stencil for the first-derivative values  of the function respectively.

This algorithm uses same matrix of weights generated from fornberg(1988) algorithm to generate new set of weights.

Corresponding tests for both cases are added, and docstrings are also updated.

I do not yet know how to use these weights for MOL/discretisation and other boundary conditions functions etc, hence have kept default . I will try to cover those next time.

### DIFF
--- a/src/derivative_operators/fornberg.jl
+++ b/src/derivative_operators/fornberg.jl
@@ -1,20 +1,27 @@
 #############################################################
-# Fornberg algorithm
+#= Fornberg algorithm
 
-# This implements the Fornberg (1988) algorithm (https://doi.org/10.1090/S0025-5718-1988-0935077-0)
-# to obtain Finite Difference weights over arbitrary points to arbitrary order.
+This implements the Fornberg (1988) algorithm (https://doi.org/10.1090/S0025-5718-1988-0935077-0)
+and hermite-based finite difference Fornberg(2020) algorithm (https://doi.org/10.1093/imanum/draa006)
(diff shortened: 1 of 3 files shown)


xtalax:
Excellent stuff, if the error characteristics from the paper hold this represents a massive step up in accuracy.

Could you please clarify how to use the outputs `d` and `e` to recover an approximation for a derivative of the specified order?

yewalenikhil65:
> or recognizing when such an equation exits within a `PDESystem` and automatically dispatching methods using these weights to these variables.
> 
Do you think something similar to [Automatic Transformation of Nth Order ODEs to 1st Order ODEs](https://mtk.sciml.ai/dev/tutorials/higher_order/) could help us ? It lowers the order by introducing new variables which are generally derivatives of original state variable !

PR #503, opened 2021-12-27
Deprecate MOL 

xtalax:
Deprecated the MOL Finite Differencing with error messages redirecting users to MethodOfLines.jl


ChrisRackauckas:
docs

PR #511, opened 2022-01-26
Complete Differences

xtalax:
Adds derivative operators that include the boundary points, fixes fornberg for order 0 (interpolations)


ChrisRackauckas:
Needs tests

PR #517, opened 2022-04-11
Nonuniform complete derivative operator constructors

xtalax:


ChrisRackauckas:
What are the "complete" operators?

xtalax:
Simply operators that also contain stencils for the 'ghost nodes', to support derivatives in the boundary conditions

ChrisRackauckas:
Add docs?

xtalax:
They're pretty much internal use only, the only methods using them are in MOL

